PR #13, opened 2020-02-01
Fix typo

AsciiWolf:


src/permissionEntryRow.ui
```diff
@@ -26,7 +26,7 @@
                 <property name="visible">True</property>
                 <property name="can_focus">False</property>
                 <property name="halign">start</property>
-                <property name="label" translatable="yes">Pemission</property>
+                <property name="label" translatable="yes">Permission</property>
```
tchx84: Thanks for catching those. Note that these strings are not user-facing, so additionally we must remove these "translatable" tags, so the string can be remove (automatically) from the .pot and .po files when I updated them.

po/es.po
```diff
@@ -57,7 +57,7 @@ msgid "Description"
 msgstr "Descripción"
 
 #: src/permissionEntryRow.ui:29 src/permissionSwitchRow.ui:29
-msgid "Pemission"
+msgid "Permission"
```
tchx84: No need to fix it manually here.

po/flatseal.pot
```diff
@@ -55,7 +55,7 @@ msgid "Description"
 msgstr ""
 
 #: src/permissionEntryRow.ui:29 src/permissionSwitchRow.ui:29
-msgid "Pemission"
+msgid "Permission"
```
tchx84: Same here. No need for this.

src/permissionSwitchRow.ui
```diff
@@ -26,7 +26,7 @@
                 <property name="visible">True</property>
                 <property name="can_focus">False</property>
                 <property name="halign">start</property>
-                <property name="label" translatable="yes">Pemission</property>
+                <property name="label" translatable="yes">Permission</property>
```
tchx84: Like mentioned above, additionally, remove the translateable tag.

tchx84 (changes requested):

src/permissionEntryRow.ui
AsciiWolf: Done. :-)

PR #14, opened 2020-02-01
Fix closing about dialog on btn click

stenosis:
I had found that the close button on the about dialog didn't worked for me. So I just looked at GTK3's JS bindings for the first time to find this bug.

### DIFF
--- a/src/application.js
+++ b/src/application.js
@@ -40,6 +40,7 @@ var FlatsealApplication = GObject.registerClass({
 
     _displayAbout() {
         const dialog = new FlatsealAboutDialog({transient_for: this._window, modal: true});
+        dialog.connect('response', dialog => dialog.close());
         dialog.present();
     }
 


tchx84:
This is strange, I can open and close the about dialog just fine.

Can you describe exactly what button you are referring to?

Can you describe the environment you are using? e.g. Linux distribution, desktop environment and flatpak version.

stenosis:
Yeah, it's odd. For all GTK or Flatpak applications, this problem only occurs with flatseal that I can't close the about dialog.

Xubuntu 18.04.3
Xfce 4.14
Flatpak 1.6.1

Flatseal

tchx84:
Here is the funny thing:

Screenshot from 2020-02-02 11-10-45

I did not add these buttons myself to the action area, and as you can see, I don't even see these. Something in your environment must be showing these. 

Since I can't reproduce this, do you mind opening the aboutDialog.ui file in Glade, and see if setting no_show_all property makes these button disappear?

Otherwise the right solution would be to explicitly hide the action area.

stenosis:
> Since I can't reproduce this, do you mind opening the aboutDialog.ui file in Glade, and see if setting no_show_all property makes these button disappear?

Sure no problem. But unfortunately the 'no_show_all' property doesn't change the behavior.
Just like you said, this could be something in my environment because the window frame of the about dialog differs from that of the main application window which makes this non-functional close button appear.

Screenshot

tchx84:
Just for the science, hah, how about removing these `<child>` nodes?

stenosis:
I thought the same so that's exactly what I did. So I recompiled the application from scretch and now it's getting crazier.. the button is still there. My guess is that this button gets inherited from Gtk.AboutDialog no matter whether this was defined in the layout file or not. But this is only a guess because this is the first GTK application that I took a closer look at and I have no experience with the GTK framework. (-:

ovari:
We are unable to click close when using Flatseal 1.3.0 when the system is in the hu (magyar - Hungarian) locale. Perhaps when the language has not be translated the button doesn't work.

image

Linux Mint 19.3 Cinnamon 64-bit

tchx84:
> We are unable to click close when using Flatseal 1.3.0 when the system is in the hu (magyar - Hungarian) locale. Perhaps when the language has not be translated the button doesn't work.
> 
> image
> 
> Linux Mint 19.3 Cinnamon 64-bit

The app itself is not adding these buttons (as shown above), and is not intended to have these buttons. I haven't had the time to reproduce and investigate this yet.

Just to make it clear, is has nothing to do with translations.

dos1:
AFAIK this is what gtk does when you force CSD to be disabled.

tchx84:
> AFAIK this is what gtk does when you force CSD to be disabled.

Hi Sebastian, do you mind elaborating a bit more?

tchx84:
This is fixed on master now, and will be available in tomorrow's release. Thanks for the help @stenosis

PR #23, opened 2020-02-08
Update hu.po

ovari:
Should
`msgid "Reset this application permissions"`
be
`msgid "Reset this application’s permissions"`

### DIFF
--- a/po/hu.po
+++ b/po/hu.po
@@ -26,7 +26,7 @@ msgstr ""
 #: data/com.github.tchx84.Flatseal.desktop.in:10
 #: data/com.github.tchx84.Flatseal.appdata.xml.in:9
 msgid "A permissions manager for Flatpak"
-msgstr ""
+msgstr "A Flatpak engedélykezelője"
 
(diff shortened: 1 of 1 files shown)


ovari:
@tchx84 can you please accept this pull request? Thank you

tchx84:
> Should
> `msgid "Reset this application permissions"`
> be
> `msgid "Reset this application’s permissions"`

Good catch but, to simplify this PR, I would recommend to just translate the current strings.

The proper fix would be to correct the string from the source and then update the .pot and po files.

ovari:
Can you please accept the hu file?

The English update can be done later.

tchx84 (approved):
Looks good. Thanks!

Ill try it locally first a bit later today and then merge it.

ovari:
Can we delete our fork?

tchx84:
> Can we delete our fork?

Sure, I already a have a local copy of your commit.

ovari:
Thanks.

Please let us know when it is on Flathub.

Issue #25, opened 2020-02-09
Electron base app in list


Eonfge:
I found that the Electron Base App is also in the list of applications. This sounds like a bit of a oversight, since there is little reason to directly control the base framework of other apps installed.

![Screenshot from 2020-02-09 18-07-11](https://user-images.githubusercontent.com/33983090/74106415-2b655f80-4b67-11ea-983a-1532fdf00294.png)

PR #27, opened 2020-02-09
#25 Implemented a blacklist for BaseApps

Eonfge:
See issue #25 where I mentioned this minor issue

### DIFF
--- a/src/model.js
+++ b/src/model.js
@@ -347,6 +347,7 @@ var FlatsealModel = GObject.registerClass({
             const appId = GLib.path_get_basename(file.get_path());
             const activePath = GLib.build_filenamev([file.get_path(), 'current', 'active']);
 
+            /* XXX will only work with app-ids that follow the .BaseApp convention */
             if (GLib.access(activePath, 0) === 0 && appId.endsWith(".BaseApp") == false)
                 list.push(appId);
 
--- a/src/model.js
+++ b/src/model.js
@@ -25,6 +25,21 @@ const _group = 'Context';
 
 const _propFlags = GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT;
 
+/*
+ * List of BaseApps can be found by searching for 'BaseApp' in the flathub project
+ * https://github.com/flathub?utf8=%E2%9C%93&q=baseApp&type=&language=
+ */
+const _blacklist = [
+    'io.elementary.BaseApp ',
+    'io.qt.qtwebkit.BaseApp',
+    'io.liri.BaseApp',
+    'io.qt.qtwebengine.BaseApp',
+    'org.electronjs.Electron.BaseApp',
+    'org.electronjs.Electron2.BaseApp',
+    'org.godotengine.godot.BaseApp',
+    'org.mozilla.Firefox.BaseApp'
+]
+
 var DELAY = 500;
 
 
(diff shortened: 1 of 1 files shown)


src/model.js
```diff
@@ -25,6 +25,21 @@ const _group = 'Context';
...
+ */
+const _blacklist = [
+    'io.elementary.BaseApp ',
+    'io.qt.qtwebkit.BaseApp',
+    'io.liri.BaseApp',
+    'io.qt.qtwebengine.BaseApp',
+    'org.electronjs.Electron.BaseApp',
+    'org.electronjs.Electron2.BaseApp',
+    'org.godotengine.godot.BaseApp',
+    'org.mozilla.Firefox.BaseApp'
```
tchx84: I don't think maintaining a hard coded list of app-ids would scale well. But, we could add a `_isBaseApp(appId)` private method to the model though, which could translate to ` return appId.endsWith('.BaseApp')`.

A bit of a hack, but I think is a decent trade-off considering `.BaseApp` is a well established convention at this point. 

EDIT: added `_` to "make" it private.

tchx84 (changes requested):

Eonfge:
I first had that, but I wish to argue that it's better to a limited blacklist of hard defined app slugs.

See, there could also be other Flatpak providers then Flathub, and they are in no way obliged to follow the same naming schematic. See [Nuvola](https://nuvola.tiliado.eu/) for example. As such, it's better to cast a net that occasionally has a false negative, then a false positive: Better not block one BaseApp from an unknown third party repository, then to block legitimate apps because string validations are inherently messy.

So, while we're add it, shall we also add `eu.tiliado.NuvolaBase` to the blacklist?

tchx84:
> I first had that, but I wish to argue that it's better to a limited blacklist of hard defined app slugs.
> 
> See, there could also be other Flatpak providers then Flathub, and they are in no way obliged to follow the same naming schematic. See [Nuvola](https://nuvola.tiliado.eu/) for example. As such, it's better to cast a net that occasionally has a false negative, then a false positive: Better not block one BaseApp from an unknown third party repository, then to block legitimate apps because string validations are inherently messy.
> 
> So, while we're add it, shall we also add `eu.tiliado.NuvolaBase` to the blacklist?

I rather see those .BaseApp in the app, which are innocuous, rather than adding the burden of maintaining a hard-coded list. You said it yourself, there could be many providers, and having a hard-coded list is just an endless mouse cat chase game.

To clarify; I am open to the idea of not-showing BaseApps, but only if we can find a way of doing it without creating bigger problems than the issue it intends to solve.

src/model.js
```diff
@@ -363,7 +347,7 @@ var FlatsealModel = GObject.registerClass({
             const appId = GLib.path_get_basename(file.get_path());
             const activePath = GLib.build_filenamev([file.get_path(), 'current', 'active']);
 
-            if (GLib.access(activePath, 0) === 0 && _blacklist.includes(appId) == false)
+            if (GLib.access(activePath, 0) === 0 && appId.endsWith(".BaseApp") == false)
```
tchx84, line 351: Can you:
1. Invert the order? Check for `.BaseApp` first. One disk access less ;)
2. Use single quotes.
3. Add a `/* XXX will only work with app-ids that follow the BaseApp convention */` (or whatever message you seem fit).

tchx84 (changes requested):
Plus, can you add a test for this new behavior? 

Basically, just add something like https://github.com/tchx84/Flatseal/tree/master/tests/content/system/flatpak/app/com.test.Example/current/active to the `tests/content` directory. Call it `com.test.BaseApp` and add a specific test in `tests/src/model.js`  to check that is the app-id is properly ignored.

tchx84:
Also, if you are not convince about this approach either,  you could research a bit more and see if there is another way to detect this condition. But if you're out of time, this will be fine for now.

Eonfge:
Added the comment. Considering the size and relevance of this feature, I think it's good enough for now.

tchx84:
I appreciate your feedback and interest in contributing to the project, but let me be clear; I can't accept contributions that won't consider such basic requirements like adding a simple test for a new behavior.

PR #29, opened 2020-02-10
l10n: Add Italian translation

MiloCasagrande:
Signed-off-by: Milo Casagrande <milo@milo.name>

### DIFF
--- a/po/it.po
+++ b/po/it.po
@@ -9,7 +9,7 @@ msgstr ""
 "Project-Id-Version: flatseal\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2020-02-06 16:37-0300\n"
-"PO-Revision-Date: 2020-02-10 21:08+0100\n"
+"PO-Revision-Date: 2020-02-13 09:12+0100\n"
 "Last-Translator: Milo Casagrande <milo@milo.name>\n"
 "Language-Team: Italian <tp@lists.linux.it>\n"
 "Language: it\n"
@@ -94,7 +94,7 @@ msgstr "Accesso al bus di sistema D-Bus (senza restrizioni)"
 
 #: src/model.js:70
 msgid "Access D-Bus session bus (unrestricted)"
-msgstr "Accesso al bus di sistema D-Bus (senza restrizioni)"
+msgstr "Accesso al bus di sessione D-Bus (senza restrizioni)"
 
 #: src/model.js:75
 msgid "Access Secure Shell agent"
--- a/po/it.po
+++ b/po/it.po
@@ -23,7 +23,7 @@ msgstr ""
 #: data/com.github.tchx84.Flatseal.desktop.in:4
 #: data/com.github.tchx84.Flatseal.appdata.xml.in:4
 msgid "Flatseal"
-msgstr "Flatseal"
+msgstr ""
 
 #: data/com.github.tchx84.Flatseal.desktop.in:10
 #: data/com.github.tchx84.Flatseal.appdata.xml.in:9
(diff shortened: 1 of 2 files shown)


tchx84 (changes requested):
This is great :+1: 

A small comment though, can you remove the translations marked with `#. TRANSLATORS: Don't translate this text`.

Thanks!

MiloCasagrande:
> A small comment though, can you remove the translations marked with `#. TRANSLATORS: Don't translate this text`.

I can, but why would I need to do that? 

I copied the original string in there and it's kind of normal (at least as GNOME translations go) to do just that, especially if you put a comment before stating to "not translate".

If a string must not be translated at all it should not be made available in the translation template file (the POT file). Or am I missing something?

tchx84:
> > A small comment though, can you remove the translations marked with `#. TRANSLATORS: Don't translate this text`.
> 
> I can, but why would I need to do that?
> 
> I copied the original string in there and it's kind of normal (at least as GNOME translations go) to do just that, especially if you put a comment before stating to "not translate".
> 
> If a string must not be translated at all it should not be made available in the translation template file (the POT file). Or am I missing something?

I see it the other way around :)

For instance, we don't have a completion % metric that we care about, there is no technical reason for copying the same string twice, and it open the possibility of introducing typos to sensitive strings like the app-id.

I need to double check whether is possible to exclude these automatically, but until then I rather have easier reviews hehe...

tchx84:
> I need to double check whether is possible to exclude these automatically

@MiloCasagrande maybe you can help me with that?

MiloCasagrande:
> @MiloCasagrande maybe you can help me with that?

On the desktop entry spec the `Icon` key is a `iconstring` value type, and I assumed it was not going to be translatable. Apparently it's not.

On the other hand, the `Name` key is a `localestring` and it will be automatically set as translatable.

The old way with `intltool` was to mark the key with an underscore to make it translatable (e.g. `_Name=`), but everybody is moving to (x)gettext now and it just takes everything that might be translatable by default.

My google-fu didn't turn up anything more useful on the subject, unfortunately.

MiloCasagrande:
> For instance, we don't have a completion % metric that we care about, there is no technical reason for copying the same string twice, and it open the possibility of introducing typos to sensitive strings like the app-id.

True, although it's easy to spot them in a PR.

Anyway, I pushed another commit to fix it in the Italian translation.

po/it.po
```diff
@@ -0,0 +1,145 @@
...
+"Flatseal is a graphical utility to review and modify basic permissions from "
+"your Flatpak applications."
+msgstr ""
+"Flatseal è uno strumento grafico per visualizzare e modificare i permessi di "
+"base delle applicazioni Flatpak."
+
+#. TRANSLATORS: Don't translate this text
+#: data/com.github.tchx84.Flatseal.appdata.xml.in:31
+msgid "Martin Abente Lahaye"
+msgstr "Martin Abente Lahaye"
```
tchx84: :point_up: :pray:

tchx84 (changes requested):

po/it.po
MiloCasagrande: Missed that one.
It's now fixed.

po/it.po
```diff
@@ -0,0 +1,145 @@
...
+msgid "Access PulseAudio sound server"
+msgstr "Accesso al server audio PulseAudio"
+
+#: src/model.js:65
+msgid "Access D-Bus system bus (unrestricted)"
+msgstr "Accesso al bus di sistema D-Bus (senza restrizioni)"
+
+#: src/model.js:70
+msgid "Access D-Bus session bus (unrestricted)"
+msgstr "Accesso al bus di sistema D-Bus (senza restrizioni)"
```
tchx84: Catched this last minute, while testing it, shouldn't it be ` sessione` ?

tchx84 (changes requested):

po/it.po
MiloCasagrande: And fixed. Thanks for catching that one.

tchx84:
Thanks, and for the research as well.

PR #34, opened 2020-02-24
Disable permissions not supported by the installed version of Flatpak

tchx84:


tchx84:
@bertob from a design POV, what do you think is more consistent? 

a) hide permissions that the user can't use (because their flatpak version is too old), or

b) disable these permissions but being explicit about it not being supported (e.g. in this PR, I set sensitivity to false to the option and add a tooltip explaining why is disabled).

I think that b) is more consistent and less confusing, e.g. if we go with a) users having the app in their phone and their laptop, with different versions of Flatpak installed, they would see different things (and that's confusing, like "why is this option showing up in my laptop but not in my phone?"). Having b) would provide them with that information.

bertob:
+1 for shaming the distro for having outdated Flatpak.

You could also go a step further and recommend a better distribution :P

tchx84:
> +1 for shaming the distro for having outdated Flatpak.
> 
> You could also go a step further and recommend a better distribution :P

didn't think of it that way, but it surely send the message haha..

PR #41, opened 2020-02-26
Enhance filesystems permissions viewer

tchx84:


tchx84:
This is still WIP:

1. I am playing around with different designs.
2. I am trying to figure out if the way I am composing these messages can be improved for i18n.
3. I am adding validation to path entry as well.

PR #66, opened 2020-04-09
model: Support custom installations paths

### LINKED ISSUE #64: Doesn't see alternative installations
Hi there, I have all of my flatpak applications installed in an alternative installation in `/xusr/flatpak`, where `/xusr` is a separate subvolume. I have granted Flatseal read permissions to `/xusr/flatpak/app` but it doesn't see any of the apps there.

tchx84:
Users should grant Flatseal access to host on Flatpak < 1.7.1 or, host-etc and the path on >= 1.7.1, to activate this.

I won't add these permissions by default, because I want Flatseal to have only minimal permissions itself.

Closes #64

### DIFF
--- a/src/model.js
+++ b/src/model.js
@@ -167,6 +167,7 @@ var FlatsealModel = GObject.registerClass({
         this._appId = '';
         this._delayedHandlerId = 0;
         this._flatpakVersion = null;
+        this._installationsPaths = null;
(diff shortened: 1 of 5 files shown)


tchx84:
@TiZ-EX1 mind helping me test this?

PR #76, opened 2020-05-16
Display application basic information

tchx84:
This is still **WIP**

Depends on https://github.com/tchx84/Flatseal/pull/75

### DIFF
--- a/com.github.tchx84.Flatseal.json
+++ b/com.github.tchx84.Flatseal.json
@@ -48,6 +48,29 @@
             ]
         },
         {
+            "name": "appstream-glib",
+            "buildsystem": "meson",
+            "config-opts": [
+		"-Ddep11=false",
+		"-Dbuilder=false",
+		"-Drpm=false",
+		"-Dalpm=false",
+		"-Dfonts=false",
+		"-Dstemmer=false",
+		"-Dman=false",
+		"-Dstemmer=false",
+		"-Dgtk-doc=false",
+		"-Dintrospection=true"
+            ],
+            "sources": [
+                {
+                    "type": "archive",
+                    "url": "https://github.com/hughsie/appstream-glib/archive/appstream_glib_0_7_17.tar.gz",
+                    "sha256": "cb612c9e634275e574fa639737cf63711358cd10b9d0d377f70025653fefdd16"
+                }
+            ]
+        },
+        {
             "name": "flatseal",
             "buildsystem": "meson",
             "sources": [
--- a/src/appInfoViewer.js
+++ b/src/appInfoViewer.js
@@ -1,0 +1,94 @@
+/* appInfoViewer.js
+ *
+ * Copyright 2020 Martin Abente Lahaye
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const {GObject, Gtk} = imports.gi;
+
+const {FlatpakApplicationsModel} = imports.models.applications;
+
+const _propFlags = GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT;
+
+const styles = {
+    NORMAL: 'normal',
+    COMPACT: 'compact',
+};
+
+var FlatsealAppInfoViewer = GObject.registerClass({
+    GTypeName: 'FlatsealAppInfoViewer',
+    Template: 'resource:///com/github/tchx84/Flatseal/appInfoViewer.ui',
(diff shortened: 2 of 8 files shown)


tchx84:
Hey @bertob ! can you design-review this one?

bertob:
I'm getting this error in Builder:

```
(com.github.tchx84.Flatseal:2): dbind-WARNING **: 16:26:33.563: AT-SPI: Error retrieving accessibility bus address: org.freedesktop.DBus.Error.ServiceUnknown: org.freedesktop.DBus.Error.ServiceUnknown
(com.github.tchx84.Flatseal:2): GLib-DEBUG: 16:26:33.603: unsetenv() is not thread-safe and should not be used after threads are created

(com.github.tchx84.Flatseal:2): Gjs-WARNING **: 16:26:35.406: JS ERROR: Error: Expected type utf8 for Argument 'str' but got type 'Object'
_init@resource:///com/github/tchx84/Flatseal/js/applicationRow.js:30:20
_setup/<@resource:///com/github/tchx84/Flatseal/js/window.js:85:25
_setup@resource:///com/github/tchx84/Flatseal/js/window.js:83:22
_init@resource:///com/github/tchx84/Flatseal/js/window.js:57:14
vfunc_activate@resource:///com/github/tchx84/Flatseal/js/application.js:75:28
main@resource:///com/github/tchx84/Flatseal/js/main.js:31:24
run@resource:///org/gnome/gjs/modules/script/package.js:222:19
```

tchx84:
> I'm getting this error in Builder:
> 
> ```
> (com.github.tchx84.Flatseal:2): Gjs-WARNING **: 16:26:35.406: JS ERROR: Error: Expected type utf8 for Argument 'str' but got type 'Object'
> ```

It seems I need to drop my attempt to parse appdata ""manually", and see if I can use appstream-glib.

A few screenshots here:

Screenshot from 2020-05-17 11-13-51

Screenshot from 2020-05-17 11-14-23

A few notes:

1. Not all applications provide all that information, e.g. don't include author or release information. For the time being, I am just using `Unknown` as default value to display. As I really don't want to make this info viewer too smart/complex.

Screenshot from 2020-05-17 11-17-24

2. I wonder if there's an existing style class in Adwaita that provides that thick shadow for the icons, for now I am using `icon-dropshadow`. But is still not thick enough, in comparison.

Screenshot from 2020-05-17 11-18-31

bertob:
That looks pretty good already.

One thing that's a bit confusing to me is the `Flatpak: 1.6.3` row. Is that the Flatpak version on the host? If so, why would we display that as a property of the app?

Using "unknown" when the data is missing seems fine to me.

tchx84:
> That looks pretty good already.
> 
> One thing that's a bit confusing to me is the `Flatpak: 1.6.3` row. Is that the Flatpak version on the host? If so, why would we display that as a property of the app?

It's the system-installed Flatpak version, and yeah, it is a bit off place... is just an important piece of information I think we should display, but I am not sure where.

> 
> Using "unknown" when the data is missing seems fine to me.

Alright.

bertob:
> It's the system-installed Flatpak version, and yeah, it is a bit off place... is just an important piece of information I think we should display, but I am not sure where.

Perhaps there could be an entry in the sidebar for the system, maybe at the top? That panel could show global information like the Flatpak version, installed SDKs, etc.

I guess the sidebar title would have to change to something other than "Applications" then.

tchx84:
> > It's the system-installed Flatpak version, and yeah, it is a bit off place... is just an important piece of information I think we should display, but I am not sure where.
> 
> Perhaps there could be an entry in the sidebar for the system, maybe at the top? That panel could show global information like the Flatpak version, installed SDKs, etc.
> 
> I guess the sidebar title would have to change to something other than "Applications" then.

Now that I refactored the "backend", I can actually move the Flatpak version to the specific tooltip where it's needed.

tchx84:
@bertob just to make sure that the switch to AppStreamGlib was worth it (it does), can you re-run this PR? Just to make sure it handles whatever was broken before? (as I can't reproduce it).

bertob:
Tried the branch, it builds fine now and looks great!

image

I'm not sure if this is a new bug, but after the leaflet folds and unfolds, the primary menu button is no longer attached to the right side of the left panel.

If possible I'd use a different date format than "year-month-day", could we do something more verbose like "May 21, 2020"? I think we have ways to localize that kind of thing nicely, depending on the locale.

In narrow mode there's not enough vertical spacing between the metadata and the first category title. I'd add e.g. an additional 12px of margin here:

image

tchx84:
> Tried the branch, it builds fine now and looks great!
> 

:muscle: 

> image
> 
> I'm not sure if this is a new bug, but after the leaflet folds and unfolds, the primary menu button is no longer attached to the right side of the left panel.
> 

Not new, I suspect something in libhandy actually, but I need to double check.

> If possible I'd use a different date format than "year-month-day", could we do something more verbose like "May 21, 2020"? I think we have ways to localize that kind of thing nicely, depending on the locale.
> 

For now ill leave it with ISO format, to keep things simple (considering how sporadic translations contributions tend to be).

> In narrow mode there's not enough vertical spacing between the metadata and the first category title. I'd add e.g. an additional 12px of margin here:
> 

Sure!

> image

bertob:
About the date format, I was told this does what we need: https://developer.gnome.org/glib/stable/glib-GDateTime.html#g-date-time-format

tchx84:
> About the date format, I was told this does what we need: https://developer.gnome.org/glib/stable/glib-GDateTime.html#g-date-time-format

That changes things :) Thanks! will give it a try.

PR #78, opened 2020-05-23
Support Show Details for applications

### LINKED ISSUE #72: [Feature Request] Show details button
It would be great to have a "Show Details" button available when there is GNOME-Software installed calling `gnome-software --details app-id` similar to what GNOME-Shell does. 
Screenshot 1
Screenshot 2

tchx84:
This could help to alleviate requests for more application management options in Flatseal.

Closes #72

### DIFF
--- a/com.github.tchx84.Flatseal.json
+++ b/com.github.tchx84.Flatseal.json
@@ -11,7 +11,8 @@
         "--socket=wayland",
         "--filesystem=/var/lib/flatpak/app:ro",
         "--filesystem=~/.local/share/flatpak/app:ro",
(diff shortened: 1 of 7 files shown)


tchx84:
This is still WIP.

@bertob what do you think ?

I personally think it looks a bit less clean in compact mode (maybe unconsciously against having something at the bottom of a scrolled window :rofl: ). But like you said, should be more ergonomic on a phone.

image

image

Maybe we need to style the action bar differently? Or is just missing , like a drop shadow.

bertob:
Looks good to me!

PR #80, opened 2020-06-06
Coexist with flatpak-override CLI changes

### LINKED ISSUE #73: Flatseal delete `persistent` overrides
Flatseal seems to delete `persistent` overrides (I'm guessing it overwrites things it doesn't support)?

```
$ flatpak override --user --show us.zoom.Zoom
[Context]
sockets=wayland;
shared=!ipc;
filesystems=!home;
persistent=.config;.zoom;
$ com.github.tchx84.Flatseal  # Do something like, turn wayland off and on again, or any other no-op
$ flatpak override --user --show us.zoom.Zoom
[Context]
sockets=wayland
shared=!ipc
filesystems=!home
```

It also seems that `flatpak --override` adds trailing `;` and Flatseal deletes them, making tracking my diffs quite a bit harder.

tchx84:
Make the permissions model aware of unsupported overrides that were made by flatpak-override and make sure the will be kept.

Closes #73

### DIFF
--- a/src/application.css
+++ b/src/application.css
@@ -189,3 +189,11 @@ separator {
 .permissions .app-info .value {
     margin-bottom: 5px;
 }
+
+.reset-button label {
+  margin-right: 5px;
+  margin-left: 5px;
+}
+.reset-button image {
+    color: @warning_color;
+}
--- a/src/com.github.tchx84.Flatseal.src.gresource.xml
+++ b/src/com.github.tchx84.Flatseal.src.gresource.xml
@@ -16,6 +16,7 @@
     <file>window.js</file>
     <file>models/applications.js</file>
     <file>models/info.js</file>
+    <file>models/unsupported.js</file>
     <file>models/permissions.js</file>
   </gresource>
 </gresources>
--- a/src/models/permissions.js
+++ b/src/models/permissions.js
@@ -20,6 +20,7 @@ const {GObject, GLib} = imports.gi;
 
 const {FlatpakInfoModel} = imports.models.info;
 const {FlatpakApplicationsModel} = imports.models.applications;
+const {FlatpakUnsupportedModel} = imports.models.unsupported;
 
 const _propFlags = GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT;
 
@@ -31,7 +32,14 @@ const _groupDescriptions = {
     filesystems: _('List of filesystem subsets available to the application'),
 };
 
-var GROUP = 'Context';
+var _groupsSupported = {
+    shared: 'Context',
+    sockets: 'Context',
+    devices: 'Context',
+    features: 'Context',
+    filesystems: 'Context',
+};
+
 var DELAY = 500;
 
 
@@ -161,7 +169,7 @@ var FlatpakPermissionsModel = GObject.registerClass({
     },
     Signals: {
         changed: {
-            param_types: [GObject.TYPE_BOOLEAN],
+            param_types: [GObject.TYPE_BOOLEAN, GObject.TYPE_BOOLEAN],
         },
     },
 }, class FlatpakPermissionsModel extends GObject.Object {
@@ -172,6 +180,7 @@ var FlatpakPermissionsModel = GObject.registerClass({
 
         this._info = new FlatpakInfoModel();
         this._applications = new FlatpakApplicationsModel();
+        this._unsupported = new FlatpakUnsupportedModel();
 
         this._notifyHandlerId = this.connect('notify', this._delayedUpdate.bind(this));
     }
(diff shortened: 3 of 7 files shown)


tchx84:
Screenshot from 2020-06-06 16-32-19

image

Hey @bertob what do you think about that warning icon? 

Context: This PR makes sure that Flatseal will not drop permissions overrides made through other tools, e.g. flatpak-override.

Scenario: But, now we have a situation where Flatseal will detect that there are changes made to an app permissions, but just not the ones you can edit through Flatseal. 

Solution: The first idea that came through my mind is to simply put that warning icon there to warn the user that there are other changes that are not visible in Flatseal, so they are aware it when resetting.

Other ideas: I considered having a dialog showing up as well... but I personally dislike dialogs.

@WhyNotHugo Want to help with some testing? :) I still need to write proper tests for this, but it seems to be working fine.

WhyNotHugo:
Sure! I'll give it a try tomorrow and give you some feedback :)

tchx84:
Updated implementation and design 

image

I think adding the extra warning as part of the existing tooltip makes more sense.

WhyNotHugo:
I went in and changed this setting:

2020-06-10T10:10:47,779318657+02:00

These are the changes for that override file:
```diff
--- a/home/.local/share/flatpak/overrides/com.valvesoftware.Steam
+++ b/home/.local/share/flatpak/overrides/com.valvesoftware.Steam
@@ -1,2 +1,3 @@
 [Context]
-filesystems=!xdg-pictures;!xdg-music;
+features=!multiarch
+filesystems=!xdg-pictures:ro;!xdg-music:ro
```

Looks like the `:ro` gets dropped.

WhyNotHugo:
2020-06-10T10:14:17,917647240+02:00

The `reset` button has no icon for me, though that may be a result of my local settings / theme (I'm using `Arc` theme, but I don't _think_ I changed anything icon related).

bertob:
@tchx84 So if I understood correctly the flow would be the same, with the only difference being the warning icon and tooltip?

I don't know how important the information being overwritten is to people, but if it is important perhaps it would be better to have a dialog detailing what is being overwritten?

tchx84:
> I went in and changed this setting:
> 
> 2020-06-10T10:10:47,779318657+02:00
> 
> These are the changes for that override file:
> 
> ```diff
> --- a/home/.local/share/flatpak/overrides/com.valvesoftware.Steam
> +++ b/home/.local/share/flatpak/overrides/com.valvesoftware.Steam
> @@ -1,2 +1,3 @@
>  [Context]
> -filesystems=!xdg-pictures;!xdg-music;
> +features=!multiarch
> +filesystems=!xdg-pictures:ro;!xdg-music:ro
> ```
> 
> Looks like the `:ro` gets dropped.

I know it messes with your diff hehe, but it's actually doing the right thing. See the following steps:

1. Check that the original Steam filesystem permissions are `xdg-pictures:ro` and `xdg-music:ro`
2. Remove these two using flatpak-override but only specifying `xdg-pictures` and `xdg-music`, which is OK because `xdg-music` takes precedence over `xdg-music:ro`.
3. Remove `multiarch` using Flatseal.
4. End up with that same diff.

Flatseal compares differences against the original permissions, while flatpak-overrides does not (it basically just writes what you passed), and that is why the `:ro` show up in the diff (because of step 1).

tchx84:
> 2020-06-10T10:14:17,917647240+02:00
> 
> The `reset` button has no icon for me, though that may be a result of my local settings / theme (I'm using `Arc` theme, but I don't _think_ I changed anything icon related).

Can you double check that you actually include an override that is not supported by Flatseal?

e.g. `flatpak override --persist=.delete_me_later com.valvesoftware.Steam`

The icon will only show-up if there are changes that are not visible (supported by) in Flatseal.

tchx84:
> @tchx84 So if I understood correctly the flow would be the same, with the only difference being the warning icon and tooltip?
> 

Yes.

> I don't know how important the information being overwritten is to people, but if it is important perhaps it would be better to have a dialog detailing what is being overwritten?

The icon (and the extra text in the tooltip) will only show up if the user has overriden something using `flatpak-override` that is not supported (and therefore, not visible in the UI) by Flatseal.

It's safe to assume that is important, but for simplicity I would just like to make it as clear as possible that the reset operation will also affect these changes.

Maybe we need to make the extra text more explicit like `, including other changes that were not made with Flatseal`, but my non-native English is already choking haha.

WhyNotHugo:
Ah, you're quite right `:ro` is fine to drop here. Makes sense to me :)

tchx84:
I will merge this as is, and improve things as we go :) Thanks for the help.

PR #81, opened 2020-06-14
Big permissions model refactor

tchx84:
The idea is to break permissions handling logic even more, so we can implement way more complex models later on without ending with spaghetti code.


tchx84:
ugh, finally!

PR #102, opened 2020-07-18
Support session and system bus names overrides

### LINKED ISSUE #84: Add --talk-name=org.freedesktop.Flatpak
The [Sandbox Permissions Reference](https://docs.flatpak.org/en/latest/sandbox-permissions-reference.html) mentions: 

`--talk-name=org.freedesktop.secrets | Talk to a named service on the session bus`

More specifically, I'd like to be able to set permissions for `--talk-name=org.freedesktop.Flatpak` Is this the same permission that is referenced above? Could you add such an option to allow users access to this? 

Linking to [corresponding issue](https://github.com/johnfactotum/foliate/issues/477) in Foliate.

tchx84:
Closes #84

### DIFF
--- a/src/application.css
+++ b/src/application.css
@@ -60,7 +60,9 @@
 .permissions .row.features ~ .features,
 .permissions .row.filesystems ~ .filesystems,
 .permissions .row.persistent ~ .persistent,
-.permissions .row.environment ~ .environment {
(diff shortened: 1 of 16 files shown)


tchx84:
I can always use help with testing ;) give it a try!

PR #112, opened 2020-07-29
widgets: Fix crash when multiple versions of libhandy are installed

mufeedali:
Prevents a crash when libhandy1 (0.80.0 and above) is installed along with libhandy 0.0 (0.0.13 and below).

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -21,6 +21,7 @@ pkg.initFormat();
 pkg.require({
     Gio: '2.0',
     Gtk: '3.0',
+    Handy: '0.0',
 });
 
 const {FlatsealApplication} = imports.application;


tchx84:
Thanks @fushinari !

You should move this to [pkg.require](https://gjs.guide/guides/gtk/gtk-tutorial/11-packaging.html#setting-dependencies-with-packaging), [here](https://github.com/tchx84/Flatseal/blob/master/src/main.js#L21-L24).

tchx84 (changes requested):
See comment above.

mufeedali:
> Thanks @fushinari !
> 
> You should move this to [pkg.require](https://gjs.guide/guides/gtk/gtk-tutorial/11-packaging.html#setting-dependencies-with-packaging), [here](https://github.com/tchx84/Flatseal/blob/master/src/main.js#L21-L24).

Thank you. I've force-pushed the needed change.

tchx84:
Thanks!

PR #132, opened 2020-08-31
window: Make menu popover easier to use on compact mode

### LINKED ISSUE #131: mobile: Menu popover too small
See https://github.com/tchx84/flathub-mobile/blob/master/README.md

tchx84:
Closes #131

### DIFF
--- a/src/application.css
+++ b/src/application.css
@@ -231,3 +231,9 @@ separator {
 .reset-button image {
     color: @warning_color;
 }
+
+/* Menu */
+
+window .menu.background label {
+    padding: 7.5px 0px;
+}


tchx84:
Hi @bertob,

I did some research over the weekend on Flathub applications on a real phone, and found some [pain points](https://github.com/tchx84/flathub-mobile/blob/master/README.md#common-pain-points) I want to address.

As a short-term solution, I am thinking to make the menu-popover items a little bigger when the app is "folded", but I am wondering if there's something more conventional... any suggestions welcome.

Thanks for all the help so far :)

EDIT: this is how it looks with this hack:

image

bertob:
I don't think it needs to be that drastic, a little bit more padding across the board (i.e. not only when it's folded) would be enough.

Android and iOS have very similar menus, and the entries aren't that much more widely spaced than ours: https://developer.apple.com/design/human-interface-guidelines/ios/controls/context-menus/

tchx84:
> I don't think it needs to be that drastic,

Drastic is my second name :laughing: 

>a little bit more padding across the board (i.e. not only when it's folded) would be enough.
 
Done.

> Android and iOS have very similar menus, and the entries aren't that much more widely spaced than ours: https://developer.apple.com/design/human-interface-guidelines/ios/controls/context-menus/

I see, interesting. Thanks for the reference.

bertob:
> >    a little bit more padding across the board (i.e. not only when it's folded) would be enough.

> Done.

Screenshot?

tchx84:
> > > a little bit more padding across the board (i.e. not only when it's folded) would be enough.
> 
> > Done.
> 
> Screenshot?

image

bertob:
Looks good :+1:

PR #136, opened 2020-09-06
Add DOCUMENTATION.md

### LINKED ISSUE #126: Better documentation page
Currently using [1], I wonder if there's a better page to replace that one.

[1] https://docs.flatpak.org/page/sandbox-permissions-reference.html

TheEvilSkeleton:
Fixes https://github.com/tchx84/Flatseal/issues/126

Here's a preview: [TheEvilSkeleton/Flatseal/DOCUMENTATION.md](https://github.com/TheEvilSkeleton/Flatseal/blob/master/DOCUMENTATION.md)


tchx84:
Ill take a look tonight, after I am done with work !

DOCUMENTATION.md
```diff
@@ -0,0 +1,116 @@
+# Documentation
+
+## Flatseal permissions
+
+This section indicates the permissions of Flatpak with better integration with Flatseal. It is based on the official [Sandbox Permissions Reference](https://docs.flatpak.org/en/latest/sandbox-permissions-reference.html).
+
+Flatseal uses the `flatpak override` command. Every change the user does will make Flatseal run the `flatpak override` command to override the default permissions that the Flatpak program originally shipped with. All overridden permissions are located inside the `~/.local/share/flatpak/overrides` and `/var/lib/flatpak/overrides` directories. If you want to read more into `flatpak override`, you can look at the [`FLATPAK OVERRIDE(1)` manpage](https://manpages.debian.org/testing/flatpak/flatpak-override.1.en.html).
```
tchx84: Both `flatseal` and `flatpak-override` command line tool use the same overrides back-end (they both create override files).

That is the reason why both share the same options.

DOCUMENTATION.md
```diff
@@ -0,0 +1,116 @@
...
+
+### Permissions
+
+#### Share
+
+List of subsystems shared with the host system.
+
+Name | Permission(s) | Type | Description
+--- | --- | --- | ---
+Network | `--share=network` and `--unshare=network` | Toggle | Permit access to the network (`--share=network`); prohibit access to the network (`--unshare=network`).
```
tchx84: It definitely helps to have the command reference but, for the last column, we could simply explain what each state of the toggle means without referencing the command.

DOCUMENTATION.md
```diff
@@ -0,0 +1,116 @@
...
+
+List of filesystem subsets available to the application.
+
+Name | Permission(s) | Type | Description
+--- | --- | --- | ---
+All filesystem files | `--filesystem=host` and `--nofilesystem=host` | Toggle | Permit read-write access to the application to the whole filesystem (excluding directories or files owned by other groups) (`--filesystem=host`); prohibit read-write acess to the application to the whole filesystem (`--nofilesystem=host`). 
+All system libraries, executables and static data | `--filesystem=host-os` and `--nofilesystem=host-os` | Toggle | [NEEDS DESCRIPTION]
+All system configurations | `--filesystem=host-etc` and `--nofilesystem=host-etc` | Toggle | [NEEDS DESCRIPTION]
+All user files | `--filesystem=home` and `--nofilesystem=home` | Toggle | Permit read-write access to the application to the user directory (`$HOME`) (`filesystem=home`); prohibit read-write access to the application to the user directory (`$HOME`) (`--nofilesystem=home`). 
+Other files | `--filesystem=[DIRECTORY]`, `--filesystem=[DIRECTORY]:ro` and `--nofilesystem=[DIRECTORY]` | Input | [NEEDS DESCRIPTION]
```
tchx84: The description here could focus on examples and what they mean, I think is easier to understand that why, what do you think ?

tchx84 (changes requested):
This is a looking great, I left some comments already! Thanks for the work so far :)

DOCUMENTATION.md
TheEvilSkeleton: Are you saying that what I wrote was false and suggesting to change it or are you explaining it further?

DOCUMENTATION.md
TheEvilSkeleton: I support your suggestion. I was thinking of the same but I had no idea how to write it. I personally think that we should give examples only to the "Other files" section because it is the only one without a toggle, which is why it needs more explanation.

I'll think of what I can do and will update it in a day or two, since I am busy with school.

DOCUMENTATION.md
TheEvilSkeleton: The only reason I reference the commands is because there are people that would want to learn the commands of Flatpak without using a GUI (me for example).

I personally think that referencing the commands is a good idea because it helps the user learn the commands in case they want to learn the commands.

I don't mind removing, though. I just want to state my opinion.

DOCUMENTATION.md
tchx84: > Are you saying that what I wrote was false and suggesting to change it or are you explaining it further?

My bad, I wasn't clear.

Flatseal doesn't use `flatpak-override` command line tool, instead, Flatseal produces the same output (override files) as the command line tool.

So that second paragraph should be corrected :)

DOCUMENTATION.md
tchx84: I think is fine to leave the references, I am just saying that, since this will be the documentation for Flatseal, it would be better to use address Flatseal options more directly, e.g. toggle on to allow access to the network, and toggle off to disallow access to the network.

DOCUMENTATION.md
tchx84: > I support your suggestion. I was thinking of the same but I had no idea how to write it. I personally think that we should give examples only to the "Other files" section because it is the only one without a toggle, which is why it needs more explanation.
> 

:+1: 

> I'll think of what I can do and will update it in a day or two, since I am busy with school.

DOCUMENTATION.md
TheEvilSkeleton: Is this good? https://github.com/tchx84/Flatseal/pull/136/commits/1663d63e51f9addc6f6bc052f20877f90738ec09

DOCUMENTATION.md
TheEvilSkeleton: I see what you mean. Unfortunately bullet-points don't work in a tables when using Markdown. That would've made everything much easier.

DOCUMENTATION.md
tchx84: Looks correct!

DOCUMENTATION.md
tchx84: It looks better now!

tchx84 (changes requested):
Looking better, what's currently missing:

1. session bus and system bus sections
2. some descriptions.

DOCUMENTATION.md
TheEvilSkeleton: Feel free to resolve the conversation

DOCUMENTATION.md
TheEvilSkeleton: Feel free to resolve the conversation

DOCUMENTATION.md
```diff
@@ -0,0 +1,121 @@
...
+
+#### Filesystem
+
+List of filesystem subsets available to the application.
+
+Name | `flatpak override` equivalent | Type | Description
+--- | --- | --- | ---
+All filesystem files | `--filesystem=host` and `--nofilesystem=host` | Toggle | Permit (`--filesystem=host`) or prohibit (`--nofilesystem=host`) read-write access to the application to the whole filesystem. If it is permitted, the application has read-write access to every file and folder owned by you, the user. The rest will be read-only.
+All system libraries, executables and static data | `--filesystem=host-os` and `--nofilesystem=host-os` | Toggle | [NEEDS DESCRIPTION]
+All system configurations | `--filesystem=host-etc` and `--nofilesystem=host-etc` | Toggle | [NEEDS DESCRIPTION]
```
TheEvilSkeleton: All the `[NEEDS DESCRIPTION]` you see between those lines are descriptions that I do not understand at all. Could you help me here @tchx84?

DOCUMENTATION.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Flatseal permissions
+
+This section indicates the permissions of Flatpak with better integration with Flatseal. It is based on the official [Sandbox Permissions Reference](https://docs.flatpak.org/en/latest/sandbox-permissions-reference.html).
+
+Flatseal follows the same concept as `flatpak-override`: overriding the default permissions that the Flatpak program originally shipped with. Every change the user does will trigger Flatseal to output the overriden permissions in the `~/.local/share/flatpak/overrides` directory.
+
+If you want to read more into Flatseal's permissions, you can look at the [Permissions section](#permissions).
+
+If you want to read more into `flatpak override`, you can look at the [`FLATPAK OVERRIDE(1)` manpage](https://manpages.debian.org/testing/flatpak/flatpak-override.1.en.html).
```
tchx84: It seems debian's man page is outdated, better use Flatpak's version https://docs.flatpak.org/en/latest/flatpak-command-reference.html#flatpak-override

DOCUMENTATION.md
```diff
@@ -0,0 +1,121 @@
...
+#### Allow
+
+List of features available to the application.
+
+Name | `flatpak override` equivalent | Type | Description
+--- | --- | --- | ---
+Development syscalls | `--allow=devel` and `--disallow=devel` | Toggle | [NEEDS DESCRIPTION]
+Programs from other architectures | `--allow=multiarch` and `--disallow=multiarch` | Toggle | [NEEDS DESCRIPTION]
+Bluetooth | `--allow=bluetooth` and `--disallow=bluetooth` | Toggle | Permit (`--allow=bluetooth`) or prohibit (`--disallow=bluetooth`) access to the application to use bluetooth.
+Controller Area Network bus | `--allow=canbus` and `--disallow=canbus` | Toggle | [NEEDS DESCRIPTION]
```
tchx84: A good start could be to re-use `flatpak-override` docs for each of those overrides. See https://docs.flatpak.org/en/latest/flatpak-command-reference.html#flatpak-override

e.g. _The multiarch feature allows the application to execute programs compiled for an ABI other than the one supported natively by the system. For example, for the x86_64 architecture, 32-bit x86 binaries will be allowed as well._

DOCUMENTATION.md
tchx84: You can find description for those:

1. man page https://docs.flatpak.org/en/latest/flatpak-command-reference.html#flatpak-override or,
2. release notes https://github.com/flatpak/flatpak/releases

e.g, from release notes,  _New filesystem permission "host-etc" and "host-os" give access to system /usr and /etc_.

DOCUMENTATION.md
```diff
@@ -0,0 +1,121 @@
...
+
+List of filesystem subsets available to the application.
+
+Name | `flatpak override` equivalent | Type | Description
+--- | --- | --- | ---
+All filesystem files | `--filesystem=host` and `--nofilesystem=host` | Toggle | Permit (`--filesystem=host`) or prohibit (`--nofilesystem=host`) read-write access to the application to the whole filesystem. If it is permitted, the application has read-write access to every file and folder owned by you, the user. The rest will be read-only.
+All system libraries, executables and static data | `--filesystem=host-os` and `--nofilesystem=host-os` | Toggle | [NEEDS DESCRIPTION]
+All system configurations | `--filesystem=host-etc` and `--nofilesystem=host-etc` | Toggle | [NEEDS DESCRIPTION]
+All user files | `--filesystem=home` and `--nofilesystem=home` | Toggle | Permit (`filesystem=home`) or prohibit (`--nofilesystem=home`) read-write access to the application to the user directory (`$HOME`).
+Other files | `--filesystem=[DIRECTORY]`, `--filesystem=[DIRECTORY]:ro` and `--nofilesystem=[DIRECTORY]` | Input | [NEEDS DESCRIPTION]
```
tchx84: For the `:ro`, `:rw` and `:create` modifiers, you can find descriptions here https://docs.flatpak.org/page/sandbox-permissions-reference.html

tchx84 (changes requested):

DOCUMENTATION.md
TheEvilSkeleton: https://github.com/tchx84/Flatseal/pull/136/commits/1d7d6b699f59f1f56cb7e7301bec3ada021ad0de

DOCUMENTATION.md
TheEvilSkeleton: https://github.com/tchx84/Flatseal/pull/136/commits/f438fda2aff7bfa0c0c9d92c8a42f050196c269f

DOCUMENTATION.md
```diff
@@ -0,0 +1,121 @@
...
+All system configurations | `--filesystem=host-etc` and `--nofilesystem=host-etc` | Toggle | Permit (`--filesystem=host-etc`) or prohibit (`--nofilesystem=host-etc`) read-write access to the application to system configurations located in `/etc`. Since this directory requires root access to write, the permission will be read-only.
+All user files | `--filesystem=home` and `--nofilesystem=home` | Toggle | Permit (`filesystem=home`) or prohibit (`--nofilesystem=home`) read-write access to the application to the user directory (`$HOME`).
+Other files | `--filesystem=[DIRECTORY]`, `--filesystem=[DIRECTORY]:ro` and `--nofilesystem=[DIRECTORY]` | Input | Permit (`--filesystem=[DIRECTORY]`) or prohibit `--nofilesystem=[DIRECTORY]` read-write access to the application to the directory you desire, for example in Flatseal, you would put `~/games` (`--filesystem=~/games` using `flatpak override`) in the box if you want read-write access to `~/games`. If you want read-only access (`ro`), then you want append `:ro`, for example in Flatseal, you would put `~/games:ro` (`--filesystem=~/games:ro` using `flatpak override`) if you want read-only access to `~/games`.
```
tchx84: Looks good :) one suggestion, how about using `PATH` instead of `DIRECTORY` ? It is shorter (making this parragh easier to read and it will fit better in the table cells) and conceptually speaking a path serves both for files and directories.

tchx84 (changes requested):
Just one suggestion for the filesystem sections, great work!

DOCUMENTATION.md
TheEvilSkeleton: It looks much better https://github.com/tchx84/Flatseal/pull/136/commits/fcba05edd39194d9374f2db2f51a8bf5f5767957

DOCUMENTATION.md
TheEvilSkeleton: That one is only for `Programs from other architectures` right?

tchx84: Correct!

TheEvilSkeleton: Multiarch: https://github.com/tchx84/Flatseal/pull/136/commits/1a3205204d9ee6da151386cb61ac8da2dcaa2fd8

DOCUMENTATION.md
```diff
@@ -0,0 +1,121 @@
...
+Shared memory | `--device=shm` and `--nodevice=shm` | Toggle | Permit (`--device=shm`) or prohibit (`--nodevice=shm`) access to the application to shared memory.
+All devices | `--device=all` and `--nodevice=all` | Toggle | Permit (`--device=all`) or prohibit (`--nodevice=all`) access to the application to all devices.
+
+#### Allow
+
+List of features available to the application.
+
+Name | `flatpak override` equivalent | Type | Description
+--- | --- | --- | ---
+Development syscalls | `--allow=devel` and `--disallow=devel` | Toggle | [NEEDS DESCRIPTION]
```
TheEvilSkeleton, line 58: Now I need to complete this

DOCUMENTATION.md
```diff
@@ -0,0 +1,121 @@
...
+Programs from other architectures | `--allow=multiarch` and `--disallow=multiarch` | Toggle | Permit (`--allow=multiarch`) or prohibit (`--disallow=multiarch`) access to the application to execute programs for an [ABI](https://en.wikipedia.org/wiki/Application_binary_interface) other than the one supported natively by the system.
+Bluetooth | `--allow=bluetooth` and `--disallow=bluetooth` | Toggle | Permit (`--allow=bluetooth`) or prohibit (`--disallow=bluetooth`) access to the application to use bluetooth.
+Controller Area Network bus | `--allow=canbus` and `--disallow=canbus` | Toggle | [NEEDS DESCRIPTION]
```
TheEvilSkeleton, line 61: And this

DOCUMENTATION.md
TheEvilSkeleton: Devel: https://github.com/tchx84/Flatseal/pull/136/commits/b4cf67d39f5986a60a82bd6d0194e7a782b00bf6

TheEvilSkeleton: Canbus: https://github.com/tchx84/Flatseal/pull/136/commits/88e3cfd0a2893544c0651694ed645e3a5a239178

DOCUMENTATION.md
```diff
@@ -0,0 +1,121 @@
...
+
+List of features available to the application.
+
+Name | `flatpak override` equivalent | Type | Description
+--- | --- | --- | ---
+Development syscalls | `--allow=devel` and `--disallow=devel` | Toggle | Permit (`--allow=devel`) or prohibit (`--disallow=devel`) access to the application to certain syscalls such as [`ptrace()`](https://en.wikipedia.org/wiki/Ptrace) and [`perf_event_open()`](https://en.wikipedia.org/wiki/Perf_(Linux)).
+Programs from other architectures | `--allow=multiarch` and `--disallow=multiarch` | Toggle | Permit (`--allow=multiarch`) or prohibit (`--disallow=multiarch`) access to the application to execute programs for an [ABI](https://en.wikipedia.org/wiki/Application_binary_interface) other than the one supported natively by the system.
+Bluetooth | `--allow=bluetooth` and `--disallow=bluetooth` | Toggle | Permit (`--allow=bluetooth`) or prohibit (`--disallow=bluetooth`) access to the application to use bluetooth.
+Controller Area Network bus | `--allow=canbus` and `--disallow=canbus` | Toggle | Permit (`--allow=canbus`) or prohibit (`--disallow=canbus`) acces to the application to use canbus sockets. You must also have network access for this to work.
```
tchx84: typo: acces (missing s)

tchx84 (changes requested):
It's getting there :)

Only missing:
* environment variables
* system and session bus permissions.

DOCUMENTATION.md
TheEvilSkeleton: Fixed https://github.com/tchx84/Flatseal/pull/136/commits/272c87cc4069f118ba9ffae8f5a6b16cbeca48a5

tchx84:
> It's getting there :)
> 
> Only missing:
> 
> * environment variables
> * system and session bus permissions.

@TheEvilSkeleton any progress with these ? :)

TheEvilSkeleton:
> > It's getting there :)
> > Only missing:
> > 
> > * environment variables
> > * system and session bus permissions.
> 
> @TheEvilSkeleton any progress with these ? :)

I still have to do those? Yikes, I think I forgot. I'll try to work with these later today or tomorrow.

Can you help me with these? I don't know how to word them.

tchx84:
> > > It's getting there :)
> > > Only missing:
> > > 
> > > * environment variables
> > > * system and session bus permissions.
> > 
> > 
> > @TheEvilSkeleton any progress with these ? :)
> 
> I still have to do those? Yikes, I think I forgot. I'll try to work with these later today or tomorrow.
> 
> Can you help me with these? I don't know how to word them.

Just start by copying what you can find in the man page and the other doc pages mentioned here, and we start from there. Don't worry about time, you almost got this, so feel free to continue when you can.

TheEvilSkeleton:
@tchx84 just pinging you to let you know that I copied what I found

tchx84:
> @tchx84 just pinging you to let you know that I copied what I found

I will check it tonight, after I done with work ;) thanks!

DOCUMENTATION.md
```diff
@@ -0,0 +1,137 @@
...
+--- | --- | --- | ---
+Files | `--persist=[FILENAME]` | Input | If the application doesn't have access to the real homedir, make the (homedir-relative) path FILENAME a bind mount to the corresponding path in the per-application directory, allowing that location to be used for persistent data. This updates the [Context] group in the metadata. This option can be used multiple times. 
+
+#### Environment
+
+List of variables exported to the application.
+
+Name | `flatpak-override` equivalent | Type | Description
+--- | --- | --- | ---
+Variables | `--env=[VAR]=[VALUE]` | Input | Set an environment variable in the application. This updates the [Environment] group in the metadata. This overrides to the Context section from the application metadata. This option can be used multiple times.
```
tchx84: I think having:

> Set an environment variable in the application.

Should be enough. The rest is a bit too technically and not strictly necessary.

DOCUMENTATION.md
```diff
@@ -0,0 +1,137 @@
...
+Other files | `--filesystem=[PATH]`, `--filesystem=[PATH]:ro` and `--nofilesystem=[PATH]` | Input | Permit (`--filesystem=[PATH]`) or prohibit `--nofilesystem=[PATH]` read-write access to the application to the directory you desire, for example in Flatseal, you would put `~/games` (`--filesystem=~/games` using `flatpak override`) in the box if you want read-write access to `~/games`. If you want read-only access (`ro`), then you want append `:ro`, for example in Flatseal, you would put `~/games:ro` (`--filesystem=~/games:ro` using `flatpak override`) if you want read-only access to `~/games`.
+
+#### Persistent
+
+List of the homedir-relative paths created in the sandbox.
+
+Name | `flatpak-override` equivalent | Type | Description
+--- | --- | --- | ---
+Files | `--persist=[FILENAME]` | Input | If the application doesn't have access to the real homedir, make the (homedir-relative) path FILENAME a bind mount to the corresponding path in the per-application directory, allowing that location to be used for persistent data. This updates the [Context] group in the metadata. This option can be used multiple times.
```
tchx84: This is one of the rare cases where the Flatpak doc is really hard to understand. What persistent does is

1. Creates a directory, relative to the home directory, but
2. Makes this directory visible only to the application, when the application is running (it doesn't really exists outside of the sandbox).
3. Makes sure that whatever is written in that directory will persistent, this means that the contents will be available next time the apps run.
3.

tchx84 (changes requested):
Left some notes :)

Plus, now only the dbus permissions are missing.

DOCUMENTATION.md
TheEvilSkeleton: https://github.com/tchx84/Flatseal/pull/136/commits/68fff6c0526d5ea819169cd9ac057342e3f04d14

DOCUMENTATION.md
TheEvilSkeleton: Doesn't persistent just move the director{y,ies} you want to the $FLATPAK_APP_ID directory, but only happens when `--nofilesystem=home`?

tchx84: The actual directory is created and does exist in ~/.var/app/ID but, when the app is running, inside the sandbox, it shows as if it was placed directly under `$HOME`.

TheEvilSkeleton: That's really intelligent and well made, but how would we write that in a way that is very easy to understand?

tchx84: haha, I suppose that is the reason why the Flatpak version of that description ins't easy to understand either...

I could be something like "Creates a persistent path under the $HOME directory, only available to the application" ?

TheEvilSkeleton: It's not really a "path under the $HOME directory" because it remaps the wanted directory to `~/.var/app/$ID/`.

tchx84: yes, but that's the kind of implementation detail that is irrelevant to someone wanting to use Flatseal. They just need to understand _what_ it does, not _how_. Because these details are so convoluted that makes the docs useless for regular users.

TheEvilSkeleton: I agree. But should we say the "what" first and then the "how", or just say the "what"?

tchx84: I think this is one of those cases where it's better to give a practical example instead of a "complete" explanation. Example:

e.g. instead of giving zoom access a the real "~/.zoom" directory, what `--pesist` does is to provide a fake "~/.zoom" directory that is only available to the application. 

something like that,

TheEvilSkeleton: The problem is that if one uses persist, some might think that `~/.zoom` will still be in use, when it reality it will be in `~/.var/app/us.zoom.Zoom/.zoom` (so the program will be at the same time, thus making the user set it up again). And even worse, persist only works when the program has no write access to the targetted directory.

This is more of an advanced feature and requires some level of knowledge instead of just saying half of the "what".

tchx84: Yeah, but we can always point to the Flatpak more complete docs. The goal should be to make Flatseal easier to use and understand first, and then we can point to more complete technical description.

Otherwise is no different from the Flatpak docs.

TheEvilSkeleton: I would say something like:

> The targeted directory will be remapped to the Flatpak program's directory (`~/.var/app/$FLATPAK_APP_ID/[DIRECTORY]`) **if it has no write access to the targeted directory**.
> 
> For example, persisting `.mozilla` will map `~/.mozilla` to `~/.var/app/org.mozilla.Firefox/.mozilla` if the Flatpak has no write access to `~/.mozilla`.

TheEvilSkeleton: > Yeah, but we can always point to the Flatpak more complete docs. The goal should be to make Flatseal easier to use and understand first, and then we can point to more complete technical description.

Fair. But it's good to make it clear because the Flatpak docs aren't clear, in my opinion.

TheEvilSkeleton: Also, what are your thoughts on https://github.com/tchx84/Flatseal/pull/136#discussion_r498863589?

tchx84: Yeah, is fine... but it's focusing on the how, not the what. It should, at least, explain why one would do such a thing (e.g. the zoom example).

TheEvilSkeleton: > Permit only to the application to have access to the targeted directory while restricting other applications from accessing it.
> 
> The targeted directory will be remapped to the Flatpak program's directory (`~/.var/app/$FLATPAK_APP_ID/[PATH]`) **if it has no write access to the targeted directory**.
> 
> For example, persisting `.mozilla` will map `~/.mozilla` to `~/.var/app/org.mozilla.Firefox/.mozilla` if the Firefox Flatpak has no write access to `~/.mozilla`.

Is it better?

Also, I'm using Firefox as an example because there are issues in Zoom with screen recording in Wayland when using `--persist=.zoom` (https://github.com/flathub/us.zoom.Zoom/pull/217#issuecomment-702905826).

tchx84: Yeah is getting there ;) and sure any example is fine.

TheEvilSkeleton: It should look good now https://github.com/tchx84/Flatseal/pull/136/commits/502f197dbfbbfbea643b59d0b36a57f6c96a5f00

TheEvilSkeleton: https://github.com/tchx84/Flatseal/pull/136/commits/fb08e338524f9957ef54a97434b4290f84020447

DOCUMENTATION.md
TheEvilSkeleton: This one is done right? Feel free to resolve it if it is.

TheEvilSkeleton:
@tchx84 do you know what is left?

There are some reviews that haven't been resolved yet. Feel free to suggest more stuff and resolve reviews that have been resolved?

tchx84:
> @tchx84 do you know what is left?
> 
> There are some reviews that haven't been resolved yet. Feel free to suggest more stuff and resolve reviews that have been resolved?

Only these last 2 permission groups are missing:

image

:)

TheEvilSkeleton:
@tchx84 I just realized: Flatpak doesn't have good docs about d-bus. It would be really cool if we can find all the d-bus (sorry for the lack of vocabulary) and explain what they do. [`org.freedesktop.portal.Fcitx`](https://github.com/flathub/com.discordapp.Discord/blob/master/com.discordapp.Discord.json#L24) is one example of a d-bus that hardly anyone knows what it means.

DOCUMENTATION.md
```diff
@@ -0,0 +1,155 @@
...
+Owns | `--system-own-name=NAME` | Input | Allow the application to own the well known name NAME on the session bus. If NAME ends with .*, it allows the application to own all matching names. This updates the [Session Bus Policy] group in the metadata. This option can be used multiple times. 
+
+#### Session Bus
+
+List of well-known names on the session bus.
+
+Name | `flatpak-override` equivalent | Type | Description
+--- | --- | --- | ---
+Talks | `--talk-name=NAME` | Input | Allow the application to talk to the well known name NAME on the session bus. If NAME ends with .*, it allows the application to talk to all matching names. This updates the [Session Bus Policy] group in the metadata. This option can be used multiple times. 
+Owns | `--own-name=NAME` | Input | Allow the application to own the well known name NAME on the session bus. If NAME ends with .*, it allows the application to own all matching names. This updates the [Session Bus Policy] group in the metadata. This option can be used multiple times.
```
tchx84: If  the user knows what Dbus is and what the session and system bus do, this works... but ideally, we could simplify this a bit by using friendlier terms like "system services the application can talk to, and that it can offer", for "talk" we can use the org,freedesktop.Notifications as an example, saying that it means it can send notifications.

tchx84 (changes requested):

DOCUMENTATION.md
TheEvilSkeleton: https://github.com/tchx84/Flatseal/pull/136/commits/a9ccbc9ded38e2a72c4213195b4c650a1268d7b3, is it good?

DOCUMENTATION.md
```diff
@@ -0,0 +1,155 @@
...
+Talks | `--system-talk-name=[NAME]` | Input | Permit the application to talk to system services. <br /> <br /> For example, inputting `org.freedesktop.Notifications` will allow the application to send notifications.
+Owns | `--system-own-name=[NAME]` | Input | Permit the application to own system services.
+
+#### Session Bus
+
+List of well-known names on the session bus.
+
+Name | `flatpak-override` equivalent | Type | Description
+--- | --- | --- | ---
+Talks | `--talk-name=[NAME]` | Input | Permit the application to talk to session services. <br /> <br /> For example, inputting `org.freedesktop.Notifications` will allow the application to send notifications.
```
tchx84: The term "inputting" is a bit weird :) We common terms like "typing down" or just "adding".

DOCUMENTATION.md
```diff
@@ -0,0 +1,155 @@
...
+Owns | `--own-name=[NAME]` | Input | Permit the application to own session services.
```
tchx84: I would "... under the given name."

tchx84 (changes requested):

DOCUMENTATION.md
TheEvilSkeleton: https://github.com/tchx84/Flatseal/pull/136/commits/ec370d26a93580054884a2716bebaa6e831d40b0

DOCUMENTATION.md
TheEvilSkeleton: https://github.com/tchx84/Flatseal/pull/136/commits/670a531e11f5a227a47d505ee99cada1efcfa314

tchx84 (changes requested):
One last request, can you change all the "Permit" terms for "Allow" instead ?

I think would be more understandable. Also, remove the WIP. I think this good enough for getting started,

TheEvilSkeleton:
Done https://github.com/tchx84/Flatseal/pull/136/commits/96546e3f666c88f73629f1714f72c18427e701e0

tchx84 (approved):
We will need to iterate on some details before I make the app point to this page, but it is a great starting point :)

Thanks for your work.

TheEvilSkeleton:
> We will need to iterate on some details before I make the app point to this page, but it is a great starting point :)
> 
> Thanks for your work.

No problem. I'll look into the docs once in a while so I can do some minor changes.

PR #144, opened 2020-09-25
applications: support custom flatpak user dir

alaviss:
Flatpak user installation directory can be moved with the `XDG_DATA_HOME`
variable (exported to flatpak as `HOST_XDG_DATA_HOME`). Check if this is
the case and set the user path accordingly.

### DIFF
--- a/com.github.tchx84.Flatseal.json
+++ b/com.github.tchx84.Flatseal.json
@@ -11,8 +11,8 @@
         "--socket=wayland",
         "--device=dri",
         "--filesystem=/var/lib/flatpak/app:ro",
-        "--filesystem=~/.local/share/flatpak/app:ro",
-        "--filesystem=~/.local/share/flatpak/overrides:create",
+        "--filesystem=xdg-data/flatpak/app:ro",
+        "--filesystem=xdg-data/flatpak/overrides:create",
         "--talk-name=org.gnome.Software"
     ],
     "cleanup": [
--- a/src/models/applications.js
+++ b/src/models/applications.js
@@ -29,8 +29,20 @@ var FlatpakApplicationsModel = GObject.registerClass({
         this._systemPath = GLib.build_filenamev([
             GLib.DIR_SEPARATOR_S, 'var', 'lib', 'flatpak',
         ]);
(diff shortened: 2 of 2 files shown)


alaviss:
It's also worth noting that this directory can also be moved with `FLATPAK_USER_DIR`, but since there's no way to program the needed permissions in I decided to leave it out. A potential solution would be to prompt the user to grant the permission themselves via the `flatpak` utility, but I don't think this is common enough to matter.

tchx84:
Hey @alaviss , thanks.

Just putting more context here, for when I get the time to review this:

> Besides standard environment variables such as XDG_DATA_DIRS and XDG_DATA_HOME, flatpak is consulting some of its own.
>
> FLATPAK_USER_DIR
> The location of the per-user installation. If this is not set, $XDG_DATA_HOME/flatpak is used.

https://docs.flatpak.org/en/latest/flatpak-command-reference.html

tchx84 (changes requested):
@alaviss, looks correct!

Double checking on Flatpak docs/source:

1. if [FLATPAK_USER_DIR](https://github.com/flatpak/flatpak/blob/master/common/flatpak-dir.c#L1546) is not set, it falls back to [XDG_DATA_HOME](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) (via [GLib.get_user_data_dir](https://gjs-docs.gnome.org/glib20~2.64.1/glib.get_user_data_dir)), which in turns falls back to `$HOME/.local/share`.

2. Once in the sandbox, [HOST_XDG_DATA_HOME](https://github.com/flatpak/flatpak/blob/master/common/flatpak-run.c#L1556) is used to access the host `XDG_DATA_HOME`.

Regarding `FLATPAK_USER_DIR` (and probably `FLATPAK_SYSTEM_DIR`), I think we should try to read it anyway. It would at least work when running outside the sandbox and when permissions to these directories _are_ given. We can check if the directory doesn't "exists", and best-effort throw a warning or something along these lines.

For what is currently implemented in this PR, just one request :)

1. s/user_data_dir/userDataDir/g

PR #146, opened 2020-09-29
Update flatpak dependencies

Eonfge:


tchx84 (approved):
Looks correct, I just happen to have issues installing org.gnome.Sdk//3.38... some random "can't find delta" issue, ill double check what's going on with that and then merge.

Thanks!

PR #156, opened 2020-10-08
Update missing po

tchx84:


tchx84:
@cho2 @stenosis I am updating these with automatic translation, if possible, I would appreciate if you can check these whenever possible.

PR #160, opened 2020-10-21
CI: add a flatpak job

bilelmoussaoui:
builds & uploads an artifact to make testing the app very simple from the user

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,20 @@
+on:
+  push:
+    branches: [master]
+  pull_request:
+name: CI
+jobs:
+  flatpak-builder:
+    name: "Flatpak Builder"
+    runs-on: ubuntu-latest
+    container:
+      image: docker.io/bilelmoussaoui/flatpak-github-actions
+      options: --privileged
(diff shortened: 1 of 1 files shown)


bilelmoussaoui:
The CI won't start until this gets merged & either a new PR is created/rebased on top of that or a new commit is pushed to master. Here's the CI in action on my fork https://github.com/bilelmoussaoui/Flatseal/pull/1/checks?check_run_id=1288177617

tchx84:
Absolute awesomeness :)

Think something like this could be also used for running the jasmine tests? see https://github.com/tchx84/Flatseal/issues/108

bilelmoussaoui:
A different job could be added yes

tchx84:
> A different job could be added yes

Greal, ill ask more details in chat ;)

Thanks for this!

bilelmoussaoui:
Actually, I've just  checked, the tests are run through meson which would require some work from the action side and it should just work later :) I don't want to tie it to meson like we do with the gitlab ci template we have at gnome to make sure other apps adopt flatpak :p 
so I will have to figure out that first

PR #173, opened 2020-10-26
Update credits

tchx84:


src/widgets/aboutDialog.ui
```diff
@@ -21,6 +21,11 @@ Kukuh Syafaat
 Åke Engelbrektson</property>
     <property name="artists">Tobias Bernard
 J.P. MacDonald</property>
+    <property name="documenters">TheEvilSkeleton</property>
```
tchx84, line 24: @TheEvilSkeleton what name should I use here?

TheEvilSkeleton: TheEvilSkeleton should be fine. Also thank you :)

PR #185, opened 2020-11-06
data: List compatible form-factors

Kekun:
This will allow the app to be available in Phosh and PureOS Store.


tchx84 (approved):
Thanks :) I was planning to do this!

PR #190, opened 2020-11-12
models: Add support for portals permissions

### LINKED ISSUE #105: support background permission
Support adding/removing background permission

tchx84:
Closes #105

NOTE: Even though this seems to work, this is only for exploratory purposes. I still need to better understand how these "dynamic" permissions work.


tchx84:
hey @bertob, I could use your advice here :)

So, I made some good progress on supporting portal-based (non-overrides) permissions, e.g. "Run in background". But I am wondering what, if any, considerations I should make from a design perspective.

Both types of permissions are fundamentally different, but I think I can make these look indistinguishable from the user's POV. My question is, is that desirable? I think is it, but I wonder if there would be any benefit from actually making a visual differentiation.

Arxcis:
Great to see progress :+1: 

For the sake of discussion about if UI should distinguish between static and dynamic, it would be nice to know, who is Flatseal's intended audience?

- As a Flatpak/Gnome-engineer, I care a lot about the difference between static and dynamic permissions.
- As a non-engineer, I don't think I even understand the difference between static and dynamic, and it would be confusing to talk to me about it. All I want is to tell the app what it can, and what it cannot do. Regardless of static or dynamic.

Hope that is helpfull :pray:

Arxcis:
May I ask, who is Flatseal's intended audience? A Gnome/Flatpak engineer may have different wishes for Flatseal, than a non-engineer. Who do you listen to, if there is a conflict of interest?

Arxcis (approved):
## Functional review
- I approve that this PR solves #105 :+1: 
- I really like that you used `Portal`, instead of `Permission`. `Portal` sounds so much cooler/less boring :+1: 
- **Suggestion:** I think the `Portal`-section should be the first section of the page, instead of the last. Since it is the most dynamic section, you may want to check it more often, than the rest of the sections.
portalmoveup


### Test: Is Slack still running?

**Given** Slack has permission to run in the background
**When** I open Slack, and click the close icon.
**Then** Slack should still be running according to `htop`

TEST PASS :+1:  :tada:
backgroundslack 

### Test: Did Slack stop running?

**Given** Slack does **not** have permission to run in the background
**When** I open Slack, and click the close icon.
**Then** Slack should stop running according to `htop`

TEST PASS :+1: :tada: 
slacknobackground

**Stretch goal:** Add `Notifications` ON/OFF as well?

tchx84:
> May I ask, who is Flatseal's intended audience? A Gnome/Flatpak engineer may have different wishes for Flatseal, than a non-engineer. Who do you listen to, if there is a conflict of interest?

Hey @Arxcis !

Thanks for your comments and your tests, like I mentioned above, this is simply a first exploratory approach and still needs more testing and cleanups.

 e.g. I need to double check if the DBus API I am using promises some forward stability/compatibility, make sure Flatseal won't break if the Dbus service is unavailable for whatever reason, write tests, and more.

Regarding your remarks, I think you're right, but I suppose it was kind of a explicit move from my part to seat Flatseal somewhere in the middle of those two type of users you describe. I want Flatseal to be "extremely" simple to use, but without hiding technical details, which I reckon is a hard thing to do.

Like I mentioned above, I prefer to blend the two types of permissions, but I won't know if it's 100% possible until I finish this properly.

tchx84:
> > May I ask, who is Flatseal's intended audience? A Gnome/Flatpak engineer may have different wishes for Flatseal, than a non-engineer. Who do you listen to, if there is a conflict of interest?
> 
> Hey @Arxcis !
> 
> Thanks for your comments and your tests, like I mentioned above, this is simply a first exploratory approach and still needs more testing and cleanups.
> 
> e.g. I need to double check if the DBus API I am using promises some forward stability/compatibility, make sure Flatseal won't break if the Dbus service is unavailable for whatever reason, write tests, and more.
> 

Good news is that Alex L. confirmed there will be forward compatibility with this API.

> Regarding your remarks, I think you're right, but I suppose it was kind of a explicit move from my part to seat Flatseal somewhere in the middle of those two type of users you describe. I want Flatseal to be "extremely" simple to use, but without hiding technical details, which I reckon is a hard thing to do.
> 
> Like I mentioned above, I prefer to blend the two types of permissions, but I won't know if it's 100% possible until I finish this properly.

tchx84:
@bertob what do you think ?

Screenshot from 2020-11-17 18-56-53

Arxcis:
#196 convinced me that a good default is to mix portal-device-permissions into the existing device-section and portal-filesystem-permissions into the existing filesystem-section.

Maybe in a separate portals-sub-section? Like the "Other files" sub-section, there could be a "Portal files" sub-section?


In the device-section, there could be a similar "Portal devices" sub-section . "Portal devices" include camera, microphone, speakers, location.

TheEvilSkeleton:
> @bertob what do you think ?
> 
> https://github.com/tchx84/Flatseal/pull/190#issuecomment-729238367

This looks amazing! The only problem is currently, a lot of flatpaks do not take advantage of portals, so the Portals category will be useless in a lot of areas.

For example, the Discord flatpak can still hear me from my microphone, despite the toggle for the Microphone portals being disabled:

image

TheEvilSkeleton:
Not entirely sure, but I changed the permissions for Pitivi, and `cat`'ing in `~/.local/share/flatpak/overrides/org.pitivi.Pitivi` doesn't show the new changes:

image

Does it mean those types of permissions are applied somewhere else?

Arxcis:
> Does it mean those types of permissions are applied somewhere else?

Yes @TheEvilSkeleton. These are dynamic permissions. They are not stored in any of the static permission files. What is the output if you run `flatpak permissions`?

**Static permissions** - You have to restart the app to apply changes
**Dynamic permissions** - Applied immediately while the app is running

TheEvilSkeleton:
> Yes. These are dynamic permissions. They are not stored in any of the static permission files. What is the output if you run `flatpak permissions`?

image

Seems to be working.

> **Static permissions** - You have to restart the app to apply changes
> **Dynamic permissions** - Applied immediately while the app is running

I never knew dynamic permissions existed. Thank you so much for being very clear with it.

Arxcis:
> I never knew dynamic permissions existed.

Dynamic permissions is not widely used yet, but more and more apps are using it. Like you said:
>The only problem is currently, a lot of flatpaks do not take advantage of portals, so the Portals category will be useless in a lot of areas.

In the future, apps will ~not require any~ **require very few** static permissions, **like wayland and network**. **For all other resources,** they will instead ask you (or your desktop environment), for what they need, when they need it, using portals (see: [xdg-desktop-portals](https://github.com/flatpak/xdg-desktop-portal)). It will take time to get to this future, because it requires apps to add support for portals. It also requires portals to be made available, for every use-case an app may have... 

 Portals store permissions you give in a dynamic permissions-storage, which you can view by running `flatpak permissions` (as demonstrated above).

bertob:
> @bertob what do you think ?

Looks good to me, though I wonder if the descriptions are needed if they're just restating the title. Perhaps they could be rephrased to add something, just removed?

tchx84:
This seems to be the proper way to test this... https://pwithnall.pages.gitlab.gnome.org/libglib-testing/docs/libglib-testing-GtDBusQueue.html

TheEvilSkeleton:
@tchx84 how do I test the "proper way"? I just looked through that page and I had no idea what I was looking at.

tchx84:
> @tchx84 how do I test the "proper way"? I just looked through that page and I had no idea what I was looking at.

I found a different way of creating unit tests for this, I will come back to this after the next release of Portfolio.

tchx84:
Still not ready for prime-time, but now that I can properly test this (as in with proper tests), is time to merge it.

TheEvilSkeleton:
@tchx84 good news!

https://user-images.githubusercontent.com/50847364/110858086-a7a90580-8287-11eb-9f3f-cb2815abeb80.mp4

tchx84:
> @tchx84 good news!
> 
>  2021-03-11.16-23-52.mp4

:clap:

PR #203, opened 2020-12-02
po: Update Czech translation

AsciiWolf:
/cc @tchx84


tchx84:
I silently updated the summary and description source strings, but re-used the existing translations. Thanks @AsciiWolf ! 

cc: @ovari @Vistaus @eson57 @cho2 @MiloCasagrande @stenosis

PR #207, opened 2020-12-12
models: Hide previously removed environment variables

### LINKED ISSUE #204: Unset environment variables don't show up/get removed
Variables set with `VAR=` get removed from the user config, and don't show up when present in the flatpak metadata.

Setting a variable to an empty string unsets it.

tchx84:
Closes #204

### DIFF
--- a/src/models/variables.js
+++ b/src/models/variables.js
@@ -116,7 +116,7 @@ var FlatpakVariablesModel = GObject.registerClass({
     }
 
     loadFromKeyFile(group, key, value, override) {
-        if (value.length === 0)
+        if (!override && value.length === 0)
             return;
 
         const dict = override ? this._overrides : this._originals;
--- a/tests/content/system/flatpak/app/com.test.Variables/current/active/metadata
+++ b/tests/content/system/flatpak/app/com.test.Variables/current/active/metadata
@@ -1,0 +1,2 @@
+[Environment]
+ENVIRONMENT=set
--- a/tests/content/user/flatpak/overrides/com.test.Variables
+++ b/tests/content/user/flatpak/overrides/com.test.Variables
@@ -1,0 +1,2 @@
+[Environment]
+ENVIRONMENT=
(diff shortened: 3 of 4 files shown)


tchx84:
Hey @YaBoiBurner

Can you test this? In case you need instructions, follow these:

```
git clone https://github.com/tchx84/Flatseal.git
cd Flatseal
git checkout fix-bug-variables
flatpak-builder --force-clean --repo=repo build com.github.tchx84.Flatseal.json
flatpak build-bundle repo flatseal.flatpak com.github.tchx84.Flatseal
flatpak install flatseal.flatpak
flatpak run com.github.tchx84.Flatseal//master
```

BurnerWah:
Doesn't seem to be working on my end.

Screenshot from 2020-12-12 16-08-55

Also, since it might be an easier test case, here's the environment section for `us.zoom.Zoom`:
```desktop
[Environment]
QT_QPA_PLATFORM=
```
Screenshot from 2020-12-12 16-12-10

~I'll hazard a guess that the regex `/^\w+=\S+$/` in `src/models/variables.js` might be causing issues, but I'm not really sure.~ changing that didn't do anything

tchx84:
Oh, I think we are talking about different things:

1) Not showing unset variables is an intended behavior, and I prefer to keep it that way.

2) What I fixed here is that, when you removed an existing variable, it would still show in the UI (even though it was unset in the overrides). But now, after you remove the variables it won't show in the UI anymore.

BurnerWah:
Ah, yeah there was definitely some confusion there.

Removing existing variables works fine.

Given this overrider for `org.qbittorrent.qBittorrent`:
```desktop
[Environment]
TMPDIR=
```
The UI goes from this:
Screenshot from 2020-12-12 16-53-33
To this:
Screenshot from 2020-12-12 16-54-42


Personally I kinda feel like there should be an option to show unset environment variables, but that's not really as important.

tchx84:
> Personally I kinda feel like there should be an option to show unset environment variables, but that's not really as important.

I understand, it's hard to keep the UI simple and cover all the cases, hehe :sweat_smile: 

In this case, my rationale is the following:

If a less technical user explicitly removes a variable (from the UI) they would expect that variable to be gone from the UI as well. The understanding that `VAR=` translates internally to unset it could be confusing for less technical users that just want to remove the variable, and it doesn't stop more technical users to just add `VAR=something-else` to override, even if it's hidden by default.

PR #211, opened 2020-12-29
Remove JACK example from shared memory

tchx84:


tchx84:
@hfiguiere please review ;)

hfiguiere:
LGTM

(can't vouch for translation though, but seems sensible to me)

PR #215, opened 2021-01-13
add polish translation

Garbulix:
I had a problem with testing the translation, because the program always loaded English translation, so I wasn't able to test it by myself.

I'm also not sure if everything is translated understandably for the end-user, so you should be ready for some polish-translation issues in the future :D

Thanks!

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -7,3 +7,4 @@ it
 nl
 pt_BR
 sv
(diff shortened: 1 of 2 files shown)


tchx84 (approved):
@Garbulix works just fine! BTW, I renamed the file to pl. Thanks!

PR #217, opened 2021-01-22
Take advantage of libhandy for styling and structure

BrainBlasted:
Makes a few changes to how Flatseal's permissions view is styled and structured, using widgets from libhandy and recommended patterns. Also makes changes to how the window is set up so we only need one leaflet, not multiple.


BrainBlasted:
## Window

Before: 

image

After:

image

## Groups

Before:

image

After:

image

tchx84:
Very impressive Chris ! Will review this ASAP.

src/widgets/window.js
```diff
@@ -258,7 +259,7 @@ var FlatsealWindow = GObject.registerClass({
             showPermissionsButton = false;
         }
 
-        const isFolded = this._headerLeaflet.folded;
+        const isFolded = this._contentLeaflet.folded;
         if (isFolded) {
             showApplicationsButton = true;
             showPermissionsButton = true;
```
tchx84, line 265: Minor issue:

BrainBlasted: I think you may have forgotten to attach an image?

tchx84: My bad I was trying to use peek to record something and accidently clicked on "single comment" :laughing:

tchx84, line 265: Minor issue:

Peek 2021-01-23 18-18


Would it be possible to disable the hovering effect when is not needed? Is a bit distracting when hovering over options like the above.

Codewise, it seems to me this should only be enabled for `FlatsealPermissionSwitchRow`.

tchx84, line 265: Minor issue:

image

For some reason, non `FlatsealPermissionSwitchRow` are a little bit bigger than they should be. Ideally, these rows should have the same height as `FlatsealPermissionSwitchRow` rows.

tchx84, line 265: Minor issue:

image

Is that always-like-is-hovering effect intended on these buttons?

tchx84, line 265: One regression:

Peek 2021-01-23 18-27

I had this [geddan-effect](https://knowyourmeme.com/memes/get-down-geddan) issue before, and I workaround it by preventing the condition that triggered it. See ad7f9541ab4e3f86eed49acde0062502d0e8282f

The root cause is that, the permissions header bar width changes _after_ the leaflet threshold has been passed. It changes because some button are hidden, and the action bar at the bottom is shown. So, it gets in this loops when the window bounces in and out the threshold.

src/widgets/window.js
```diff
@@ -106,7 +104,7 @@ var FlatsealWindow = GObject.registerClass({
 
         this._layoutNotifyId = 0;
         this._updateControlsPlacement();
-        this._headerLeaflet.connect('notify::folded', this._updateControlsPlacement.bind(this));
+        this._contentLeaflet.connect('notify::folded', this._updateControlsPlacement.bind(this));
```
tchx84, line 107: Question:

Is there a more "handistic" way to update the controls buttons placement? I would love to get rid of  `_updateControlsPlacement`

tchx84 (changes requested):
Hey Chris,

This looks really good :) So much code gone and looks generally better.

I left comments for 3 minor issues, 1 _ugly_ regression and 1 question.

Thanks a lot!

src/widgets/window.js
BrainBlasted: HdyHeaderGroup

src/widgets/window.js
BrainBlasted: Yes

BrainBlasted: That's the .list-button style class

BrainBlasted: Screenshot from 2021-01-23 14-07-46
How is this?

tchx84: hard to tell by eye, maybe check with inspector? just to make sure it has the same height as the `FlatsealPermissionSwitchRow` :)

This is what I am doing currently https://github.com/tchx84/Flatseal/blob/master/src/widgets/permissionSwitchRow.ui#L11

BrainBlasted: They do :)

tchx84:
hey @BrainBlasted , any progress on this? :)

AsciiWolf:
Wow, this looks really great! :-)

tchx84:
> Wow, this looks really great! :-)

it does, @BrainBlasted can you push whatever latest version you have ? So, someone else might pick it up :)

tchx84:
> > Wow, this looks really great! :-)
> 
> it does, @BrainBlasted can you push whatever latest version you have ? So, someone else might pick it up :)

I will give a try on finishing this on a new PR, so all this good effort doesn't go to waste :)

PR #219, opened 2021-01-26
window: Fix search by application name

### LINKED ISSUE #218: Search in sidepane is limited to package ID
### Usecase
The Element app (https://flathub.org/apps/details/im.riot.Riot) has the package ID `im.riot.Riot` and the appname `Element`. 
Searching for "Element" will list 0 flatpaks on Flatseal 1.6.7.

### Description
Searching in the sidepane will filter flatpaks by their package ID but not their app name. As the app name is shown in Bold letters in Flatseal it's quite confusing for the user why it's not listed when used in a Search.

tchx84:
Closes #218

### DIFF
--- a/src/widgets/window.js
+++ b/src/widgets/window.js
@@ -220,7 +220,12 @@ var FlatsealWindow = GObject.registerClass({
         if (text.length === 0)
             return true;
 
-        return row.appId.toLowerCase().includes(text.toLowerCase());
+        const subString = text.toLowerCase();
+
+        return (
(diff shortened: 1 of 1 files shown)


tchx84:
@Mershl trivial fix but, can you help me test it?

Mershl:
Thanks @tchx84! Changes look good. Just tested it locally with every usecase I could came up with and it works as expected.

tchx84:
> Thanks @tchx84! Changes look good. Just tested it locally with every usecase I could came up with and it works as expected.

Thank you!

PR #239, opened 2021-03-23
DOCUMENTATION: Add Portals permissions descriptions

tchx84:


DOCUMENTATION.md
```diff
@@ -99,6 +100,20 @@ Name | Type | Description | `flatpak override` equivalent
 Talks | Input | Allow the application to talk to session services. <br /> <br /> For example, adding `org.freedesktop.Notifications` will allow the application to send notifications. | `--talk-name=[NAME]`
 Owns | Input | Allow the application to own session services under the given name. | `--own-name=[NAME]`
 
+
+### Portals
+
+Portals, also know as dynamic permissions, are granted by users during runtime.
```
TheEvilSkeleton:
```suggestion
[Portals](https://github.com/flatpak/flatpak/wiki/Portals), also known as dynamic permissions, are permissions that require manual user intervention to grant or deny access to specific permissions.

Dynamic permissions take priority over static permissions.

For example, denying the application to access the webcam using static permissions (`--nodevice=all`) but allowing it to access the webcam using portals will ignore the static permission as Flatpak prioritizes portals, thus allowing the application to access the webcam.
```

DOCUMENTATION.md
```diff
@@ -99,6 +100,20 @@ Name | Type | Description | `flatpak override` equivalent
...
+
+Name | Type | Description | Portal
+--- | --- | --- | ---
+Background | Toggle | Allow the application to run in the background | `org.freedesktop.portal.Background`
+Notifications | Toggle | Allow the application to send notifications | `org.freedesktop.portal.Notification`
+Microphone | Toggle | Allow the the application to listen to your microphone | `org.freedesktop.portal.Device`
+Speakers | Toggle | Allow the application to play sounds to your speakers | `org.freedesktop.portal.Device`
+Camera | Toggle | Allow the application to record videos with your camera | `org.freedesktop.portal.Device`
+Location | Toggle | Allow the application to access your location | `org.freedesktop.portal.Location`
```
TheEvilSkeleton:
```suggestion
Background | Toggle | Allow the application to run in the background. | `org.freedesktop.portal.Background`
Notifications | Toggle | Allow the application to send notifications. | `org.freedesktop.portal.Notification`
Microphone | Toggle | Allow the the application to listen to your microphone. | `org.freedesktop.portal.Device`
Speakers | Toggle | Allow the application to play sounds to your speakers. | `org.freedesktop.portal.Device`
Camera | Toggle | Allow the application to record videos with your camera. | `org.freedesktop.portal.Device`
Location | Toggle | Allow the application to access your location. | `org.freedesktop.portal.Location`
```

The rest of the descriptions have a period at the end. It might be nitpicky, but it's for consistency.

TheEvilSkeleton (changes requested):
I have some stuff to suggest.

DOCUMENTATION.md
tchx84: Well, I think we will need to find a middle ground :laughing:  because now is way too technical, e.g. it really not necessary to explain `processes that run separately from the application`, What we need to communicate here is that these are permissions that do not come by default and that is up to the user to allow these accesses  or not.

tchx84: Another thing that we need to include there somehow, is that these take precedence over static permissions.

TheEvilSkeleton: > Another thing that we need to include there somehow, is that these take precedence over static permissions.

Can you elaborate on this? I edited my suggestion, but I failed to write the last sentence properly.

tchx84: We need to, somehow, explain that, if you revoke webcam access by removing `--devices=all` static permission, but later allow access though devices portal, the portals will take higher priority and allow the application to access the webcam.

This a pretty common confusion among Flatpak users, especially when talking about filesystem, e.g. hey I removed all filesystem permissions in Flatseal, but the application can still access this file I opened last week.

That happens because the access to that file was done through the document portals, and that takes higher priority over static permissions.

TheEvilSkeleton: I think I think I figured it out. What do you think?

tchx84: > allocate specific data to an application without needing to access a range of permissions

This part is still way too technical, we really need something for regular people like "Portals are not granted by default, and require users to explicitly allow these accesses from the application".

> For example, disabling all devices using static permissions (`--nodevice=all`) but enabling device access through portals will ignore the static permission as it prioritizes portals, thus giving the application access to devices.

This is pretty close, but I would use something more specific like the webcam to make it easier to grasp.

TheEvilSkeleton: Just updated. What do you think?

TheEvilSkeleton (approved):

TheEvilSkeleton:
@tchx84 do you think we should add a warning or something to tell the user to not use portals? I think portals should be recommended to be only used for experienced users.

tchx84:
> @tchx84 do you think we should add a warning or something to tell the user to not use portals? I think portals should be recommended to be only used for experienced users.

From user's POV, I don't think these are any riskier than static permissions TBH.

PR #262, opened 2021-04-15
portals: Support wallpaper

tchx84:


tchx84:
Hi @matthiasclasen,

I am adding Portals support to Flatseal, case by case, and there's something I noticed with this one I need to understand. I am wondering if you can help me out here :)

If an application tries to use this wallpaper portal, but the permission store [finds](https://github.com/flatpak/xdg-desktop-portal/blob/master/src/wallpaper.c#L158-L164) it's disallowed, will the portal simply ignore the request? I just need to understand what [this line](https://github.com/flatpak/xdg-desktop-portal/blob/master/src/wallpaper.c#L162) implies.

Thanks in advance!

Issue #125, opened 2020-08-13
Search permissions


tchx84:
Taken from a  comment in Flatseal gnome-software page

> Only thing missing is a permission search feature

Flatseal currently supports 30 different permissions. Maybe something like [this](https://twitter.com/alexm_gnome/status/1288503112496484355/photo/2) could be of help.

@bertob by any chance, do you know of other examples were search is implemented among settings / options? So I can based this on.

bertob:
Libhandy's HdyPreferencesWindow search is probably a good example to look at

tchx84:
@YaLTeR please feel free to describe your design there :)

YaLTeR:
So my issue is that:
- when opening Flatseal I want to type to search for the app
- when the app is open it would be nice to type to search for permissions

I initially proposed focusing the app search entry by default on launch, however on mobile this brings up the OSK which is awkward.

Therefore my idea is to add an initial empty state with no app selected (currently on launch the first app is selected). When typing on the empty state it will type to search for the app, and when typing on an app permissions page it will type to search for the permissions.

tchx84:
> Therefore my idea is to add an initial empty state with no app selected (currently on launch the first app is selected). When typing on the empty state it will type to search for the app, and when typing on an app permissions page it will type to search for the permissions.

The difficulty I see is that both, the app list and permissions list, "exist" in the exact same place. I am no designer, but wouldn't this feel a bit weird? I mean, to have the exact same action (typing) behaving differently in the exact the same place (the same view/window)?

bertob:
I'd take a step back here and figure out what use cases wer're trying to address here first. Is it about searching for a specific permission within the page of an app? Or searching for the value of the same permission across all apps? Something else?

tchx84:
> I'd take a step back here and figure out what use cases wer're trying to address here first. Is it about searching for a specific permission within the page of an app? Or searching for the value of the same permission across all apps? Something else?

good point, I am assuming the first case (within the page of the currently selected app).

bertob:
In that case a second search button/bar in the content pane would be the most obvious solution, no?

tchx84:
> In that case a second search button/bar in the content pane would be the most obvious solution, no?

This would work for mobile fine but, honest question, for the desktop, wouldn't be confusing to have two search buttons on the same header bar ?

I am quite pragmatic, so if you think that doesn't break some sacred design vow, I wouldn't mind* hah

bertob:
[...]

PR #276, opened 2021-04-26
window: Set proper accelerator for activating search

### LINKED ISSUE #274: Typing after launch no longer searches for apps
I might be misremembering but I'm fairly sure before 1.7.0 typing right after launch would search for the app. Either way this is not the case on 1.7.0, and I'd like to request this because the first thing you do after opening Flatseal is find the app you need.

tchx84:
Closes #274

### DIFF
--- a/src/widgets/docsViewer.ui
+++ b/src/widgets/docsViewer.ui
@@ -81,6 +81,7 @@
                     <property name="icon-name">system-search-symbolic</property>
                   </object>
                 </child>
+                <accelerator key="f" signal="clicked" modifiers="GDK_CONTROL_MASK"/>
               </object>
               <packing>
                 <property name="position">1</property>
@@ -149,6 +150,7 @@
                         <property name="icon-name">go-down-symbolic</property>
                       </object>
                     </child>
+                    <accelerator key="g" signal="clicked" modifiers="GDK_CONTROL_MASK"/>
                     <style>
                       <class name="next-button"/>
                     </style>
--- a/src/widgets/window.ui
+++ b/src/widgets/window.ui
@@ -58,6 +58,7 @@
                         <property name="icon-name">system-search-symbolic</property>
                       </object>
                     </child>
+                    <accelerator key="f" signal="clicked" modifiers="GDK_CONTROL_MASK"/>
                   </object>
                   <packing>
                     <property name="position">1</property>


tchx84:
@YaLTeR I prefer not to activate the search entry by default, so I think an extra ctrl+f could be a good-enough compromise.

What do you think?

YaLTeR:
Hm, why not activate it by default? I think type to search is one of the common patterns in GNOME (e.g. if you open Settings and type you'll start searching). One argument I see is that type to search should search preferences within the currently selected app, which is why I'm suggesting only focusing app search once at launch—when you want to use it. Perhaps even start with an empty state rather than selecting the first app in the list at launch, this way it's clear that typing will search for apps (since there are no preferences visible).

tchx84:
> Hm, why not activate it by default?

If I activate it by default, then I need to focus it on default (or it wouldn't make much sense, as it would still require one step to use it), and by focusing by default creates this scenario where I launch Flatseal on my phone, the virtual keyboard takes half of my screen when I just want to browse for an app that would be much easier to find if I had that screen space available.

>I think type to search is one of the common patterns in GNOME (e.g. if you open Settings and type you'll start searching). 

I haven't looked at detail, but I suspected there could be a difference between apps that don't have free-form text entries, at the top-level window, e.g. settings, and those that do, e.g. like Flatseal. But maybe there's a simple approach to this? Without having to mess with keyboard events that later creates another issue.

Would you be interested in looking into it?

Also, note that `ctrl+f` is quite a common pattern as well.

> One argument I see is that type to search should search preferences within the currently selected app, which is why I'm suggesting only focusing app search once at launch.

I haven't really thought about this (also see my comment above), but have you considered how this could work on a phone?

YaLTeR:
Oh, hmm, yeah, I don't have any experience in GTK development for phones.

Maybe then empty initial state + type to search which searches apps on the empty state and preferences on non-empty state will work?

tchx84:
> Oh, hmm, yeah, I don't have any experience in GTK development for phones.
> 
> Maybe then empty initial state + type to search which searches apps on the empty state and preferences on non-empty state will work?

Well, I think that we need to tackle the issues separately, and and then adapt the UI when a good design emerges. Let's continue the "search permissions" topic here https://github.com/tchx84/Flatseal/issues/125 (which has been opened for a while now, mostly due for the lack of clarity on the design side of things).

As for this PR, it addresses the ability of initiating search using the keyboard with the widely common `ctrl+f` accelerator. So ill merge this for now.

PR #316, opened 2021-09-07
pathRow: Set !(negation) prefix as valid

### LINKED ISSUE #314: Flatseal shows valid filesystem settings as invalid
Telegram has recently added a setting like ```!~/.TelegramDesktop``` but Flatseal shows it as an invalid option where it should be something valid. 

Also see: https://github.com/flathub/org.telegram.desktop/issues/310#issuecomment-911491628

Screenshot from 2021-09-02 12-54-35

tchx84:
Closes #314

### DIFF
--- a/src/models/filesystemsOther.js
+++ b/src/models/filesystemsOther.js
@@ -119,7 +119,8 @@ var FlatpakFilesystemsOtherModel = GObject.registerClass({
             .filter(p => !this.constructor.isOverriden(this._overrides, p));
 
         const overrides = [...this._overrides]
-            .filter(p => !this.constructor.isNegated(p));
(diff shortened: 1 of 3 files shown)


tchx84:
@xscd @gasinvein can you help me test this? try to break it :pray:

gasinvein:
Seems to behave properly, i.e. removing the negated entry adds the un-negated entry to overrides, and adding it back removes the override.

gasinvein:
Hmm, a proper tooltip for such entries seems missing, currently the tolltip says `Can modify and read:`

tchx84:
> Hmm, a proper tooltip for such entries seems missing, currently the tolltip says `Can modify and read:`

good catch!

tchx84:
I will address the tooltip issue with a separate PR.

PR #340, opened 2021-10-15
applicationRow: Use HdyActionRow

BrainBlasted:
HdyActionRow gives us title and subtitle styling for
list rows that's consistent with other GNOME apps.

The previous styling used italics for subtitles,
which is general discouraged. See: https://developer.gnome.org/hig/guidelines/typography.html

### DIFF
--- a/src/application.css
+++ b/src/application.css
@@ -6,21 +6,6 @@
     margin: 10px;
 }
 
-.applications .icon {
-   padding-left: 5px;
-}
-
-.applications .name,
-.applications .app-id {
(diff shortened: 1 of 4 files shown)


tchx84:
I see a lot of red in that diff, thanks Chris :) Ill test this during the weekend.

BrainBlasted:
Added a commit to adjust icon size.

tchx84 (approved):

tchx84:
This looks pretty good! Thanks!

PR #341, opened 2021-10-17
application: Use the standard shortcut to quit

BrainBlasted:
The standard shortcut for quitting applications is
Ctrl+Q, not Alt+Q.


tchx84 (changes requested):
this looks good Chris, can you also do the same change in the applications.css? Which is used for the docs viewer.

BrainBlasted:
Fixed.

tchx84:
Thanks!

PR #342, opened 2021-10-17
Add --allow=per-app-dev-shm

Arxcis:
## Why?

From Flatpak 1.11.1 [release notes](https://github.com/flatpak/flatpak/releases/tag/1.11.1):
>- Instances of the same app-ID can optionally share their /dev/shm directory
(enabled by a new --allow flag, --allow=per-app-dev-shm)

## TODO 
- [x] Build and run Flatseal. 
    Here is what I did:
    ```sh
    git clone https://github.com/tchx84/Flatseal.git
    cd Flatseal/
    sudo apt install flatpak-builder
    flatpak install org.gnome.Sdk/x86_64/41
    flatpak install org.gnome.Platform/x86_64/41
    flatpak-builder --force-clean --repo=repo/ build/ com.github.tchx84.Flatseal.json
    flatpak build-bundle repo/ Flatseal.flatpak com.github.tchx84.Flatseal
    flatpak install --user Flatseal.flatpak
    flatpak run --branch=master com.github.tchx84.Flatseal
    ```
- [x] Test 

| State | In GUI | In overrides file |
|-----------------|---------------|-----------|
| Toggled on |  image | image |
|  Toggled off | image | image |

### DIFF
--- a/src/models/features.js
+++ b/src/models/features.js
@@ -62,7 +62,7 @@ var FlatpakFeaturesModel = GObject.registerClass({
             },
             'features-per-app-dev-shm': {
                 supported: this._info.supports('1.11.1'),
-                description: _('Per app /dev/shm'),
+                description: _('Application Shared Memory'),
                 option: 'per-app-dev-shm',
                 value: this.constructor.getDefault(),
                 example: 'allow=per-app-dev-shm',
--- a/src/models/features.js
+++ b/src/models/features.js
@@ -60,6 +60,13 @@ var FlatpakFeaturesModel = GObject.registerClass({
                 value: this.constructor.getDefault(),
                 example: 'allow=canbus',
             },
+            'features-per-app-dev-shm': {
+                supported: this._info.supports('1.11.1'),
+                description: _('Per app /dev/shm'),
+                option: 'per-app-dev-shm',
+                value: this.constructor.getDefault(),
+                example: 'allow=per-app-dev-shm',
+            },
         };
     }
 
--- a/tests/src/testModels.js
+++ b/tests/src/testModels.js
@@ -32,7 +32,7 @@ const {
 
 setup();
 
-const _totalPermissions = 36;
+const _totalPermissions = 37;
 
 const _basicAppId = 'com.test.Basic';
 const _oldAppId = 'com.test.Old';
--- a/tests/src/testModels.js
+++ b/tests/src/testModels.js
@@ -33,7 +33,6 @@ const {
 setup();
 
 const _totalPermissions = 37;
-const _stablePermissions = _totalPermissions - 7;
 
 const _basicAppId = 'com.test.Basic';
 const _oldAppId = 'com.test.Old';
@@ -557,7 +556,7 @@ describe('Model', function() {
         permissions.appId = _basicAppId;
         const total = permissions.getAll().filter(p => p.supported).length;
 
-        expect(total).toEqual(_stablePermissions);
+        expect(total).toEqual(_totalPermissions - 7);
     });
 
     it('handles missing .flatpak-info', function() {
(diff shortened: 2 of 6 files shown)


tests/src/testModels.js
```diff
@@ -157,6 +157,7 @@ describe('Model', function() {
         expect(permissions.features_devel).toBe(true);
         expect(permissions.features_multiarch).toBe(true);
         expect(permissions.features_canbus).toBe(true);
+        expect(permissions.features_per_app_dev_shm).toBe(true);
```
tchx84, line 160: For this to actually work, you need to add this permission to the test permission files. Check which appId is used by this test to find the proper file.

src/models/features.js
```diff
@@ -60,6 +60,13 @@ var FlatpakFeaturesModel = GObject.registerClass({
                 value: this.constructor.getDefault(),
                 example: 'allow=canbus',
             },
+            'features-per-app-dev-shm': {
+                supported: this._info.supports('1.11.1'),
+                description: _('Per app /dev/shm'),
```
tchx84: Shared Memory per Application?

tchx84 (changes requested):
hey @Arxcis ! Thanks for working on this, just a few minor comments.

tests/src/testModels.js
Arxcis: Maybe fixed in fe451a1

src/models/features.js
Arxcis: Maybe fixed in fe451a1

Arxcis: Yes. Good suggestion :+1:  Since you asked, I have a few more in my head :sweat_smile: :

**Suggestions** :thinking: 
- Shared Memory per Application (credit to @tchx84)
- Shared Memory by Application
- Per Application Shared Memory 
- Shared Memory between Application instances
- Application instances share memory
- Application scoped Shared Memory
- Application specific Shared Memory
- Application owned Shared Memory
- Application local Shared Memory
- Application Shared Memory           

The last one here is my favorite, because it is similar to "Host Shared Memory", which is a concept I already understand :exclamation:

`(Host|Application) Shared Memory`:
- `(Host|Application)` - refers to the scoping of the Shared Memory. 
    - `Host` scope - Means the shared memory is accessible from anywhere on the host system, and from Flatpaks which have opened the hole --device=shm. 
    - `Application` scope - Means the shared memory is only accessible from instances of the same application.
- `Shared Memory` - refers to `/dev/shm`
- Example usage: 
    >"I have put the pipe in `Application Shared Memory`, so it is accessible from any instance of my application". 
    >"Oh, you should have put the pipe in `Host Shared Memory` instead, because I want easy access to it on the host."

Arxcis: Fix in f363a39 and ecf7faa. See screenshot in description :camera:

tchx84:
@Arxcis  You need to increase some counters in the test files :)

Arxcis:
> @Arxcis You need to increase some counters in the test files :) 

Done in 4e55ea1 :sweat_smile:

tchx84:
There's still one counter off ! It seems.

Arxcis:
> There's still one counter off ! It seems.

Sorry. I don't know how to run the tests locally before I push. Fix here 88740f9

tchx84:
This might give you some hints for running it locally https://github.com/tchx84/gtk-apps-testing-docker-action/blob/gjs/Dockerfile

tchx84 (changes requested):
Tests still failing...

Arxcis:
> This might give you some hints for running it locally https://github.com/tchx84/gtk-apps-testing-docker-action/blob/gjs/Dockerfile

Ok. Thank you. I will look into it👍

tchx84:
This is looking great now, thanks @Arxcis :)

PR #347, opened 2021-10-22
Misc updates

tchx84:


tchx84:
@Arxcis is this correct?

src/widgets/aboutDialog.ui
```diff
@@ -33,7 +33,8 @@ Mufeed Ali
 Leorize
 Christopher Davis
 Lctrs
-Will Thompson</property>
+Will Thompson
+Jonas Solsvik</property>
```
Arxcis: Thank you very much @tchx84 🙏 It is an honor to be included in this list 😁

Arxcis: I would prefer if you add my middle name as well 👍 
```suggestion
Jonas Johan Solsvik</property>
```

tchx84: Of course!

PR #364, opened 2021-11-04
Updated Dutch translation

Vistaus:
🙂

### DIFF
--- a/po/nl.po
+++ b/po/nl.po
@@ -332,7 +332,7 @@ msgstr "_Documentatie"
 
 #: src/widgets/menu.ui:15
 msgid "_Keyboard Shortcuts"
-msgstr "_Sneltoetsen"
+msgstr "Snel_toetsen"
 
 #: src/widgets/menu.ui:19
 msgid "_About Flatseal"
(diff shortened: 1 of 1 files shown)


po/nl.po
```diff
@@ -325,23 +323,20 @@ msgid "No software manager found"
...
 #: src/widgets/menu.ui:11
-#, fuzzy
 msgid "_Documentation"
-msgstr "Documentatie"
+msgstr "_Documentatie"
 
 #: src/widgets/menu.ui:15
 msgid "_Keyboard Shortcuts"
-msgstr ""
+msgstr "_Sneltoetsen"
```
tchx84: I think you're already marking an `s` in `Detail_s tonen`, so I think it would be better to mark another character here, e.g. `S_neltoetsen` ?

tchx84 (changes requested):

po/nl.po
Vistaus: You're right, my bad - should've catched that one! Anyway, I've corrected it just now, thanks! 🙂

tchx84 (approved):

tchx84:
ty ty!

Vistaus:
yw yw!

PR #373, opened 2021-11-14
window: Fix keyboard navigation on selecting applications

tchx84:


tchx84:
note to self: need to force scroll to the top...

Arxcis:
@tchx84 Awesome work! Thank you thank you :pray:

PR #376, opened 2021-11-15
fix: overriding negated permissions does not always display correctly

Arxcis:
## Why?
To fix the bug demonstrated by #375.

## Changelog
- Add fix in `models/shared.js` (see: https://github.com/tchx84/Flatseal/pull/376/files#diff-d0299c66c0a3b34e060b5057b9ee15ad7372dc8b8c04b1d8d7f26224a0ae00a4L106-L107)
- Add test  'Load negate permissions'
- Add test 'Load negate overrides'


tchx84:
Can clarify this a bit?

If I understand correctly the scenario seems to be the following;

1. originals permissions include something like `!~/folder`.
2. we add an override with `~/folder`
3. The resulting permission ends up wrong?

If this is correct, I am trying to reproduce this with the telegram flatpak...

1. It originally has `!~/.TelegramDesktop`
2. I add an override `~/.TelegramDesktop`
3. The override file seems correct, and Flatseal shows `~/.TelegramDesktop` correctly.

Arxcis:
Please try another original permission like `!host` and `!home`, then override with `home` and `host` and see if it still shows up correctly.

**UPDATED with more details**
Expect the overrides file to be correct, but Flatseal ignoring overrides when displaying permissions. Remember to RESTART Flatseal, after adding the overrides, because before you restart, the overrides will show correctly if you added the overrides from the GUI. It is only AFTER restart they will not show correctly anymore.

tests/src/testModels.js
```diff
@@ -135,7 +137,7 @@ describe('Model', function() {
         expect(appIds).not.toContain(_baseAppId);
     });
 
-    it('loads permissions', function() {
+    it('loads positive permissions', function() {
```
tchx84: No need for this change.

tests/src/testModels.js
```diff
@@ -171,7 +173,7 @@ describe('Model', function() {
         expect(permissions.variables).toEqual('TEST=yes');
     });
 
-    it('loads overrides', function() {
+    it('loads negative overrides', function() {
```
tchx84: How about using the term "negated", instead.

tchx84 (changes requested):
Last note, do you know of any real app that presents this issue? (that has a negated permission, and is not telegram?)

tchx84:
> Please try another original permission like `!host` and `!home`, then override with `home` and `host` and see if it still shows up correctly.
> 
> **UPDATED with more details** Expect the overrides file to be correct, but Flatseal ignoring overrides when displaying permissions. Remember to RESTART Flatseal, after adding the overrides, because before you restart, the overrides will show correctly if you added the overrides from the GUI. It is only AFTER restart they will not show correctly anymore.

Another critical comment is that, the changes you're making, won't don't have any effect on other permissions models, like "other filesystem", because you're only modifying the "shared" permission model. Not all models subclass "shared" and some override the `updateProxyProperty` method.

So please double check.

Arxcis:
> Last note, do you know of any real app that presents this issue? (that has a negated permission, and is not telegram?)

## 3 examples
I found 3 examples of negative permissions from crawling the Flatstat flathub-file -> https://raw.githubusercontent.com/Arxcis/flatstat/master/data/flathub/metafiles.json

| APP ID | Negative permissions | URL | Expect issue? |
|-------------------------|----------------------------|----|----|
| org.signal.Signal  | --nodevice=shm, --nodevice=kvm | https://github.com/flathub/org.signal.Signal/blob/master/org.signal.Signal.json#L18-L19 |  :x: Yes |
| org.shotcut.Shotcut | --nosocket=wayland | https://github.com/flathub/org.shotcut.Shotcut/blob/master/org.shotcut.Shotcut.yaml#L11 |  :x: Yes |
| org.telegram.desktop | --nofilesystem=~/.TelegramDesktop | https://github.com/flathub/org.telegram.desktop/blob/beta/org.telegram.desktop.yml#L27 |  :green_circle:  No   |
 

## Demo: org.signal.Signal

Let's take `org.signal.Signal` as an example, which has the following negative permissions:
```
        "--nodevice=kvm",
        "--nodevice=shm",
```

Here is what happens when I try to toggle on `kvm` and `shm`:

| Before toggle | After toggle | After restart |
|--------------------|---------------|-------------------|
| image | image | image |

Arxcis:
> Last note, do you know of any real app that presents this issue? (that has a negated permission, and is not telegram?)

By the way. I entirely forgot it, but this question is trivial to answer in [Flatstat](https://arxcis.github.io/flatstat/) :smile: 
image

tests/src/testModels.js
Arxcis: Fixed in 602a2dd

tests/src/testModels.js
Arxcis: Fixed in 602a2dd

tchx84 (changes requested):
I finally had the time to review this, looks good. The change make sense. 

Only 1 change, forgive my insistence, please use the term "negate**d**", to keep it consistent.

tchx84 (approved):

tchx84:
Great work!

tchx84:
Just in time for the release this week ;)

PR #398, opened 2022-02-17
Opt-in to Dark mode user preference

A6GibKm:
See https://gitlab.gnome.org/GNOME/Initiatives/-/issues/32

### DIFF
--- a/com.github.tchx84.Flatseal.json
+++ b/com.github.tchx84.Flatseal.json
@@ -1,7 +1,7 @@
 {
     "app-id": "com.github.tchx84.Flatseal",
     "runtime": "org.gnome.Platform",
-    "runtime-version": "41",
+    "runtime-version": "42",
     "sdk": "org.gnome.Sdk",
     "separate-locales": false,
     "command": "com.github.tchx84.Flatseal",
--- a/src/application.js
+++ b/src/application.js
@@ -108,5 +108,8 @@ var FlatsealApplication = GObject.registerClass({
         Handy.init();
         this._setupActions();
         this._setupStylesheet();
+
+        Handy.StyleManager.get_default().set_color_scheme(
+            Handy.ColorScheme.PREFER_LIGHT);
     }
 });
--- a/src/main.js
+++ b/src/main.js
@@ -24,6 +24,7 @@ pkg.require({
     Gio: '2.0',
     Gtk: '3.0',
     Handy: '1',
+    WebKit2: '4.0',
 });
 
 const {FlatsealApplication} = imports.application;
--- a/src/meson.build
+++ b/src/meson.build
@@ -1,6 +1,8 @@
 pkgdatadir = join_paths(get_option('datadir'), meson.project_name())
 gnome = import('gnome')
 
+dependency('libhandy-1', version : '>= 1.5')
+
 src_res = gnome.compile_resources('com.github.tchx84.Flatseal.src',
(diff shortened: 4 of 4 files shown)


com.github.tchx84.Flatseal.json
```diff
@@ -1,7 +1,7 @@
 {
     "app-id": "com.github.tchx84.Flatseal",
     "runtime": "org.gnome.Platform",
-    "runtime-version": "41",
+    "runtime-version": "master",
```
tchx84: For this, we will need to wait for GNOME 42...

src/widgets/docsViewer.js
```diff
@@ -18,6 +18,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+imports.gi.versions.WebKit2 = '4.0';
```
tchx84: Curious, is this related?

tchx84 (changes requested):

com.github.tchx84.Flatseal.json
A6GibKm: Ill update it either when libhandy 1.6 is released or when GNOME 42 is released.

src/widgets/docsViewer.js
A6GibKm: Yes, needed for the master runtime, otherwise WebKit will default to GTK 4, ideally we just bundle libhandy 1.6 and there is no need to use the `master` runtime.

com.github.tchx84.Flatseal.json
A6GibKm: I added libhandy 1.5.90 to the manifest rather than setting it to master.

src/widgets/docsViewer.js
A6GibKm: This is no longer needed.

A6GibKm:
mmm CI is failing.

Btw there is a workflow for flatpaks, see https://github.com/bilelmoussaoui/ashpd/blob/master/.github/workflows/demo-ci.yml for example.

tchx84:
> mmm CI is failing.
> 
> Btw there is a workflow for flatpaks, see https://github.com/bilelmoussaoui/ashpd/blob/master/.github/workflows/demo-ci.yml for example.

I am aware, the problem is... Flatseal's portal tests rely on DBus services (provided by the test suite itself) that won't run with Bilal's tool.

See https://github.com/bilelmoussaoui/flatpak-github-actions/issues/25 (Ill try to find time this week and see if I can make it work)

A6GibKm:
Updated so to use the 42 runtime which bundles libhandy. Sadly it will be years before libhandy 1.6 reaches ubuntu/debian repos.

tchx84:
@A6GibKm as a workaround, we could install libhandy 1.6 here https://github.com/tchx84/gtk-apps-testing-docker-action/tree/gjs

A6GibKm:
Sounds good to me, but please take care of it, CI is the bane of my existence.

tchx84:
I pushed the changes to my CI image, but... it seems I can't re-trigger the job myself, it simply stays at "queued" forever.

tchx84:
Still nothing..

tchx84:
@A6GibKm I created a new PR to see if I an revive this. See https://github.com/tchx84/Flatseal/pull/416

PR #400, opened 2022-02-17
docsViewer: Make dialog modal

A6GibKm:


src/widgets/docsViewer.js
```diff
@@ -54,6 +54,8 @@ var FlatsealDocsViewer = GObject.registerClass({
         const [width, height] = parent.get_size();
         this.default_width = width;
         this.default_height = height;
+        this.modal = true;
```
tchx84, line 57: Hmm, I think this is not behaving as expected.

The windows are not getting "attached" (e.g. on GNOME press the super key and you will noticed it), and I can still focus the main window (which glitches the docsViewer window position :dizzy_face: ).

I suspect it happens because it's a `Handy.ApplicationWindow` not a dialog, e.g. a few hints must be missing?

tchx84 (changes requested):

src/widgets/docsViewer.js
A6GibKm: Ill test it later, but no, there shouldn't be any difference between a GtkDialog or a HdyWindo.

A6GibKm: I tested and modal works alright here. I am not sure whats happening.

tchx84 (approved):

PR #406, opened 2022-03-10
pathRow: Add :reset mode

### LINKED ISSUE #402: Remove `:ro` suffix when overriding existing read-only permissions
Hello,

First of all, thanks for this excellent app! Very helpful to spot and overly broad permissions in many Flatpak apps.

I am running Flatseal 1.7.5 from Flatpak. I noticed that when removing permission to existing paths with read-only (`:ro`) access, Flatseal preserves the `:ro` suffix. For example, when removing `home:ro` permission for an app in Flatseal, `!home:ro` is added in its override file.

For instance, you can try removing `home:ro` permission from `fi.skyjake.Lagrange` or `org.telegram.desktop`. The `flatpak` CLI will emit a warning:

```
$ flatpak list
F: Filesystem suffix "ro" is not applicable for --nofilesystem
```

Removing the `:ro` suffix manually from the override file suppresses this warning.

tchx84:
Closes #402

### DIFF
--- a/src/models/filesystems.js
+++ b/src/models/filesystems.js
@@ -1,4 +1,4 @@
-/* exported FlatpakFilesystemsModel */
+/* exported FlatpakFilesystemsModel getDefault */
 
 /* filesystems.js
  *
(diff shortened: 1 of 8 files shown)


tchx84:
@SaswatPadhi can you give this a try?

SaswatPadhi:
Hi @tchx84,

I just built Flatseal from this branch and it looks like `:ro` suffix is not generated anymore when removing access.

Thanks for the quick fix!!

PR #416, opened 2022-03-29
Switch to Handy.StyleManager

tchx84:


tchx84:
@A6GibKm thanks :)

A6GibKm:
Thanks, note that the 42 runtime was released a while ago, it would be good to set that in the flatpak manifest and remove libhandy if it is already present in the manifest.

tchx84:
Switching to 42 was done in a previous commit :)

PR #425, opened 2022-05-02
Highlight permissions changed by the user or globally

### LINKED ISSUE #279: Highlight changed permissions
Moving this idea out of https://github.com/tchx84/Flatseal/issues/163 to reduce confusion.

The idea is to somehow highlight permissions rows that were changed (compared to the application defaults), e.g using:

* border
* background-color
*  badge
* etc

@bertob any ideas?

tchx84:
With this, Flatseal will not only show the current set of granted permissions but will also show which of those permissions are overrides and where these overrides came from, e.g. from the user or from global overrides.

Closes #279

### DIFF
--- a/src/application.css
+++ b/src/application.css
@@ -49,6 +49,13 @@ row .content .path {
     background-color: @theme_bg_color;
 }
 
+row .status {
+    color: @theme_selected_bg_color;
+}
+row.activatable .status {
+    margin-right: 8px;
+}
+
 row .content .bus entry,
 row .content .variable entry,
 row .content .path entry {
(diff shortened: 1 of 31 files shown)


tchx84:
@bertob what do you think? 

Option 1 (edit icon to the right):

image

When the user hovers the icon, it will display a tooltip specifying where the override came from, e.g. by the user or globally
image

Same applies to fine-grained permissions like filesystem, variables, etc.
image

tchx84:
@nahuelwexd what do you think?

Option 2 (alert icon to the left):

image

image

nyx-lyb3ra:
I personally think the icon should go to the right, we just need to find a better icon than the pencil.

PR #429, opened 2022-05-07
models: Support editing global overrides

### LINKED ISSUE #210: Global permission overrides
Is it in the scope of Flatseal that it could also allow managing global permission overrides?

I'm talking about an additional list item called "All applications" (e.g. placed above the search bar), which would display a simplified version of the generic application details view, but containing just the global permission settings (perhaps including a clearly visible warning that global permission overrides have less priority than application-specific permission overrides).

tchx84:
Closes #210


tchx84:
No design, no nothing. Ideas welcome.

PR #450, opened 2022-06-04
i18n: Add Occitan language

Mejans:
First partially translated version to start with.


po/oc.po
```diff
@@ -0,0 +1,578 @@
...
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n > 1);\n"
+"X-Generator: Poedit 3.0.1\n"
+
+#. TRANSLATORS: Don't translate this text
+#: data/com.github.tchx84.Flatseal.desktop.in:4
+#: data/com.github.tchx84.Flatseal.appdata.xml.in:4
+msgid "Flatseal"
+msgstr "Flatseal"
```
rusty-snake, line 25: > TRANSLATORS: Don't translate this text

Mejans: Depends what you call untranslated...it's the same string so it's untranslated, or we should fill in the corresponding msgstr here, thus I am wondering where is there a translation key.

rusty-snake: Good question, looks like this is done a bit inconsistent when I look at the other translations.

Mejans: ufff, I was asleep when I wrote xD Where=>why

po/LINGUAS
```diff
@@ -1,17 +1,18 @@
...
 de
 es
+fr
 hu
 id
 it
 nl
-pt_BR
-sv
+oc
```
tchx84, line 12: if it's not much asking, could you fix the sorting in one commit and add the oc entry together with the second commit? Once the file actually exists.

tchx84 (changes requested):

po/LINGUAS
Mejans: I don't how to do this =/

rusty-snake: `git rebase -i [--autostash --autosquash] $(git merge-base HEAD master)` and the `edit` the commit. See *SPLITTING COMMITS* in `man 1 git-rebase`.

Mejans: I did everything via the user interface ^^
If I clone my fork and run this line, will it be enough? I'm really not into git :S

rusty-snake: > If I clone my fork and run this line, will it be enough?

No.

Mejans: :(

PR #454, opened 2022-06-10
Save only valid D-Bus Names

JakobDev:
Flatpak has a Bug that Apps didn't start when you give them Permission to talk to a invalid D-Dbus name (flatpak/flatpak#4940). Until this is fixed, it is better to block writing invalid D-Bus names. If someone does accidental use a invalid D-Bus name for global and ignore the warning, no Flatpak App will start until the problem is solved with the Flatpak cli or the overrides file is deleted.

### DIFF
--- a/src/models/sessionBus.js
+++ b/src/models/sessionBus.js
@@ -23,12 +23,15 @@ const {GObject} = imports.gi;
 const {FlatpakSharedModel} = imports.models.shared;
 const {FlatsealOverrideStatus} = imports.models.overrideStatus;
 
+/* https://dbus.freedesktop.org/doc/dbus-specification.html */
+const EXP = /^(([A-Z]|[a-z]|[0-9]|_|-)+)(\.(([A-Z]|[a-z]|[0-9]|_|-)+))+(\.\*){0,1}$/;
 
 var FlatpakSessionBusModel = GObject.registerClass({
     GTypeName: 'FlatpakSessionBusModel',
 }, class FlatpakSessionBusModel extends FlatpakSharedModel {
     _init() {
         super._init({});
+        this._expression = new RegExp(EXP);
     }
 
     getPermissions() {
@@ -180,7 +183,9 @@ var FlatpakSessionBusModel = GObject.registerClass({
     saveToKeyFile(keyFile) {
         const group = this.constructor.getGroup();
         Object.entries(this._overrides).forEach(([key, value]) => {
-            keyFile.set_value(group, key, value);
+            if (this._expression.test(key)) {
+                keyFile.set_value(group, key, value);
+            }
         });
     }
 


src/models/sessionBus.js
```diff
@@ -23,12 +23,15 @@ const {GObject} = imports.gi;
 const {FlatpakSharedModel} = imports.models.shared;
 const {FlatsealOverrideStatus} = imports.models.overrideStatus;
 
+/* https://dbus.freedesktop.org/doc/dbus-specification.html */
+const EXP = /^(([A-Z]|[a-z]|[0-9]|_|-)+)(\.(([A-Z]|[a-z]|[0-9]|_|-)+))+(\.\*){0,1}$/;
```
rusty-snake, line 27: - hyphens are not allowed
- regex can be simplified (`([A-Z]|[a-z])` -> `([A-Za-z])`)
- names must not start with an digit

```suggestion
const EXP = /^([A-Za-z_][A-Za-z0-9_]+)(\.[A-Za-z_][A-Za-z0-9_]+)+(\.\*)?$/;
```

(untested suggestion)

JakobDev: This RegEx is not from me. It is already used in Flatseal.
https://github.com/tchx84/Flatseal/blob/562ddec1b12c2a41c64e08daedf7576681f41345/src/widgets/busNameRow.js#L27-L28
It is used to display the valid icon in the D-Bus edit field. If it is wrong, it should be fixed in both places.

src/models/sessionBus.js
```diff
@@ -180,7 +183,9 @@ var FlatpakSessionBusModel = GObject.registerClass({
     saveToKeyFile(keyFile) {
         const group = this.constructor.getGroup();
         Object.entries(this._overrides).forEach(([key, value]) => {
-            keyFile.set_value(group, key, value);
+            if (this._expression.test(key)) {
```
tchx84, line 186: For consistency, instead of doing this here, can you move this to `updateFromProxyProperty`? Like I do it here https://github.com/tchx84/Flatseal/blob/master/src/models/variables.js#L86

src/models/sessionBus.js
tchx84, line 27: To make this better than I currently do it, how about we officially move the definition of this esoteric regexp to this model (e.g. `var BUS_NAME_REGEX = ...`), and then we import this constant here https://github.com/tchx84/Flatseal/blob/master/src/widgets/busNameRow.js#L28 (instead of duplicating it)

tchx84 (changes requested):

src/models/sessionBus.js
A6GibKm: hypens *are* allowed, but just on the last segment of the name

PR #460, opened 2022-07-04
allow spaces in env vars value

### LINKED ISSUE #459: Space in env vars not allowed
I would like to add spaces to the value of the environement variables
eg: `CHROMIUM_USER_FLAGS="--enable-features=UseOzonePlatform --ozone-platform=wayland"`
But the input verification does not allow me to do that.

gmorer:
Closes #459

### DIFF
--- a/src/models/variables.js
+++ b/src/models/variables.js
@@ -23,6 +23,7 @@ const {GObject} = imports.gi;
 const {FlatpakSharedModel} = imports.models.shared;
 const {FlatsealOverrideStatus} = imports.models.overrideStatus;
 
+var VAR_REGEXP = /^\w+=[\S ]+$/;
 
 var FlatpakVariablesModel = GObject.registerClass({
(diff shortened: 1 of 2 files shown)


src/models/variables.js
```diff
@@ -164,6 +164,6 @@ var FlatpakVariablesModel = GObject.registerClass({
         this._overrides = {};
         this._globals = {};
         this._originals = {};
-        this._expression = new RegExp(/^\w+=\S+$/);
+        this._expression = new RegExp(/^\w+=[\S ]+$/);
```
tchx84: Can you can help me clean this up a bit by moving that regexp definition to a `var VAR_REGEXP` on the model and the just import from the widget, so we don't have 2 definitions?

tchx84 (changes requested):

tchx84 (approved):

tchx84:
Thanks!

Issue #418, opened 2022-04-02
Port Flatseal to use the Gnome 42 Platform 


MikuOcta:
It would be nice for the app to be ported to Platform 42 (and possibly to libadwaita) for more consistency with the other apps

rusty-snake:
https://github.com/tchx84/Flatseal/blob/1b9774d92ed54accd2f4e7c84bf61d0f9d276142/com.github.tchx84.Flatseal.json#L3-L4

MikuOcta:
but what about libadwaita port? Libadwaita will give a fresh look and a performance boost

tchx84:
Maybe rename the request to "port to gtk4 / libadwaita" ? re. that It won't happen for the next release, but I really want to make it happen this year (intentions).

ghost:
> https://github.com/tchx84/Flatseal/blob/1b9774d92ed54accd2f4e7c84bf61d0f9d276142/com.github.tchx84.Flatseal.json#L3-L4

And when this commit will hit flathub ? (not a troll i do not know the release time).

ghost:
@tchx84 i ask again to see if there is any news ? because your app is the last one i use to be on 41 all other have switched to 42

rusty-snake:
Flatseal 1.8.0 with org.gnome.Platform//42 is out.

tchx84:
I am closing this since this seems to be more about moving to 42 runtime. Let's leave the port of libadwaita in a new issue.

PR #464, opened 2022-08-03
GTK4/libadwaita port

fxtressia:
Hiiiii!!! <3

This PR solves #418's author's request for a GTK4 port

But at the moment, it's not yet complete. ~~The Flatseal window will be completely empty for now!!!~~ Fixed nowww!!!
- [x] <details><summary>Fix widgets not showing up!</summary> 
image
It seems many widgets' `visible` properties are set false. Turning them on manually trough the GTK inspector results in this screenshot!!!
image
 
</details>

- [x] Text boxes won't allow writing!!
   - [x] The search text box
   - [ ] The preferences text boxes
- [ ] Switch to an alternative for webkitGTK. WebkitGTK for GTK4 is not ready yet for now!!! But considering this is just for the help menu, we could just open the HTML file trough Xdg Open?


fxtressia:
image
Yayy!!!

tchx84 (changes requested):
@lovelyyfiaaa woah!! <3

Ok, OK, this is going to take a few iterations... and to make this digestible for me *please* :pray: 

1. Rebase on top of current master.
2. Let's leave linting preferences for later, and make this PR exclusively about GTK4/Adw. Nothing more nothing less.

I really appreciate this :heart_eyes:

fxtressia:
> Let's leave linting preferences for later, and make this PR exclusively about GTK4/Adw. Nothing more nothing less.


Sorry, but apparently my code editor was configured to automatically format things on save!!!

fxtressia:
@tchx84 Can I ask you some questions?

1. There seems to be an "actionBar" that wasn't visible in the GTK3 Flatseal, but it exists!
   In my port it was visible, though I removed them because it wasn't intended to be there in the GTK3 version, it seemed. Should I remove them or enable them?
2. Emmm, since WebkitGTK for GTK4 isn't ready yet, would you want the Documentation button to open the HTML file in the default browser instead

tchx84:
> @tchx84 Can I ask you some questions?
> 

Absolutely, please keep asking.

> 1. There seems to be an "actionBar" that wasn't visible in the GTK3 Flatseal, but it exists!
>    In my port it was visible, though I removed them because it wasn't intended to be there in the GTK3 version, it seemed. Should I remove them or enable them?

It's should only be visible when the window's width shrinks pass a threshold (try it on the GTK3 version to see the behavior).

> 2. Emmm, since WebkitGTK for GTK4 isn't ready yet, would you want the Documentation button to open the HTML file in the default browser instead

Given that constraint, that's probably a fair approach.

Let's not aim for perfect yet, let's just see far this can go.

fxtressia:
> It's should only be visible when the window's width shrinks pass a threshold (try it on the GTK3 version to see the behavior).

Ah, I see!! 
Screenshot of a portrait view of Flatseal showing the settings for Darkbar

JanDeDinoMan:
> Text boxes won't allow writing!! 

After a bit of investigating this is because in src/widgets/window.ui all widgets have the property can-focus set to False. Setting these to True or removing the the property (since they default to True) will resolve this problem. As seen here: https://github.com/JanDeDinoMan/Flatseal/commit/ca6a44f45fdc7c347f53b070ee19d2cbd76f0a46

fxtressia:
> > Text boxes won't allow writing!!
> 
> After a bit of investigating this is because in src/widgets/window.ui all widgets have the property can-focus set to False. Setting these to True or removing the the property (since they default to True) will resolve this problem. As seen here: [JanDeDinoMan@ca6a44f](https://github.com/JanDeDinoMan/Flatseal/commit/ca6a44f45fdc7c347f53b070ee19d2cbd76f0a46)

Okayyy

Sorry, I haven't been able to work on this again for quite some time since I went busy. I'll start working again on this PR.

A6GibKm:
These buttons should be flat

image

The reason this happens is because they are setting a GtkImage as their child instead of using the button `icon-name` property.

A6GibKm:
This GtkListBox

image

Should use the `sidebar` css class.

A6GibKm:
The reset button 

image

should be on the right of the headerbar, one has to use `<child type="end">` to achieve this.

A6GibKm:
Padding and positions are off, compare

image

image

also some text requires the `dim-label` class here. Note that the current way of setting the font type is via css classes, for example the title can use `title-1` or `title-2`, and the subtitle can use the `heading` class. You can see the available classes with https://flathub.org/apps/details/org.gnome.design.Typography

A6GibKm:
Every button in a row should use the `flat` css class

image

note that the margins on the bottom of the rows do not coincide with the margin on the top. This would be easier to acomplish with AdwActionRow.

A6GibKm:
This button should be flat and have a 6px separation with the switch. Note that toggling the switch twice won't hide the "clear" button. I dont know if the later is an issue in stable.

image

A6GibKm:
The about dialog is broken for some reason

image

Please use AdwAboutDialog instead. I think the issue is that the dialog is a subclass of GtkAboutDialog which is final, i.e. not subclassable.

A6GibKm:
The controllers appear twice in "mobile mode" , and the "back" button appears incorrectly as insensitive.

image

the controls on the bottom don't hide themselves when you increase the size of the window too.

A6GibKm:
The search bar requires some tweaks

image

Ideally it should be a GtkSearchBar containing a GtkSearchEntry, this handles input (which is not working correctly), right now it is using a revealer.

See https://docs.gtk.org/gtk4/method.SearchBar.set_key_capture_widget.html.

A6GibKm:
The environment variables looks odd

image

I am not quite sure how to fix it, maybe using AdwEntryRow? This requires design I think.

A6GibKm:
I see uses of `            <property name="icon_size">6</property>`, this takes an enum https://docs.gtk.org/gtk4/enum.IconSize.html instead of numerical values.

A6GibKm:
I don't know if this is supported in gtk4

```
@binding-set WindowBindings {
  bind "<Primary>q" { "close" () };
  bind "<Primary>f" { "find" () };
}

window {
  -gtk-key-bindings: WindowBindings;
}
```
Please use https://docs.gtk.org/gtk4/method.Application.set_accels_for_action.html

A6GibKm:
The enviaroment variables throw an error 

```
(com.github.tchx84.Flatseal:2): Gjs-CRITICAL **: 20:50:01.914: JS ERROR: TypeError: row.destroy is not a function
_remove@resource:///com/github/tchx84/Flatseal/js/widgets/pathsViewer.js:56:11
_remove@resource:///com/github/tchx84/Flatseal/js/widgets/variableRow.js:67:14
main@resource:///com/github/tchx84/Flatseal/js/main.js:38:24
run@resource:///org/gnome/gjs/modules/script/package.js:206:19
```

when adding/removing them.

A6GibKm:
I see some errors of the like
```
(com.github.tchx84.Flatseal:2): Gtk-WARNING **: 20:51:43.759: Finalizing GtkInspectorLogs 0x7f9c78004150, but it still has children left:
 - GtkBox...
```

Every widget that uses `set_parent()` should use `unparent()` in its dispose vfunc.  I think this actually comes from setting a child widget in the UI file to a widget that does not support this.

A6GibKm:
Despite the large number of issues, the port looks very good and promising. Thanks! Ill do a more careful review of the code once it leaves Draft status.

tchx84:
> Sorry, I haven't been able to work on this again for quite some time since I went busy. I'll start working again on this PR.

@lovelyyfiaaa  No worries, glad you can look at this again. @A6GibKm did a great first-pass review so I think there's a clear way forward.

A6GibKm:
Regarding the focusing issues, I would remove all uses of `can-focus` and `focusable` in the codebase.

For example

```
  <template class="FlatsealWindow" parent="AdwApplicationWindow">
    <property name="can-focus">False</property>
```

will be passed down to children of the window, completely screwing up focus everywhere.

A6GibKm:
The classes GtkRevealer and GtkAboutDialog are final meaning that they cannot be subclassed, note that `FlatsealUndoPopup` subclasses GtkRevealer.

A6GibKm:
The window has a minimal width of 364px, this won't work in mobile, it should reach 360px.

fxtressia:
> I don't know if this is supported in gtk4
> 
> ```
> @binding-set WindowBindings {
>   bind "<Primary>q" { "close" () };
>   bind "<Primary>f" { "find" () };
> }
> 
> window {
>   -gtk-key-bindings: WindowBindings;
> }
> ```
> Please use https://docs.gtk.org/gtk4/method.Application.set_accels_for_action.html

Yup, they're GTK3 APIs, and it seems it was  removed.

A6GibKm:
The css variable `@theme_selected_bg_color` is probably not defined, see https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/named-colors.html, maybe this should use the accent color?.

A6GibKm:
I am getting this error

```
(com.github.tchx84.Flatseal:2): Gtk-WARNING **: 22:36:18.362: Failed to set text 'Todoist: To-Do List & Tasks' from markup due to error parsing markup: Error on line 1: Entity did not end with a semicolon; most likely you used an ampersand character without intending to start an entity — escape ampersand as &amp;
```

it probably comes from https://github.com/flathub/com.todoist.Todoist/blob/master/com.todoist.Todoist.metainfo.xml#L6, but idk what string in the app reads this.

fxtressia:
> css

Okayy.. Also, would we want sidebar or navigation sidebar? Navigation sidebar feels awesome, but I'm not sure what others would think.

fxtressia:
@A6GibKm Also, I'm getting this issue where Flatseal actually doesn't use the `styles.css` file. Could you help investigate? Thanks.

fxtressia:
> The controllers appear twice in "mobile mode" , and the "back" button appears incorrectly as insensitive.
> 
> image
> 
> the controls on the bottom don't hide themselves when you increase the size of the window too.

Yeah, that's a regression that I am aware for quite some time now.

A6GibKm:
> @A6GibKm Also, I'm getting this issue where Flatseal actually doesn't use the `styles.css` file. Could you help investigate? Thanks.

This bit here
```js
var FlatsealApplication = GObject.registerClass({
    GTypeName: 'FlatsealApplication',
}, class FlatsealApplication extends Gtk.Application {
    _init() {
        super._init({
            application_id: 'com.github.tchx84.Flatseal',
            flags: Gio.ApplicationFlags.FLAGS_NONE,
        });

        this._window = null;

    }
```
it should extend Adw.Application this will auto load RESOURCE_BASE_PATH/style.css, so bellow the `flags` line add `resource_base_path: '/com/github/tchx84/Flatseal/',`.

src/widgets/pathsViewer.js
```diff
@@ -53,7 +53,9 @@ var FlatsealPathsViewer = GObject.registerClass(
 
     _remove(row) {
       this._box.remove(row);
-      row.destroy();
+      if (row.destroy) {
+        row.destroy();
```
A6GibKm: Note that `destroy` is not a thing in gtk4 (except with windows).

What are you trying to acomplish here? Remove a row from a listbox? In such a case there is gtk_list_box_remove (or gtk_box_remove).

tchx84: that's from the original code, it's basically resetting the widget to re-generate it from the data coming from the model (e.g. list of "other files").

A6GibKm: I just saw that you are already removiing the row correctly. There is no need for the destroy then, GTK gets rids of anything which goes out of scope.

tchx84: Yeah, I know, but in practice, it wasn't really doing that (it was leaking somehow), and it did require this extra step.

A6GibKm: Then someone is holding a reference to the row, probably a struct held it as one of its fields, idk. As I said previously `destroy` is not a thing anymore and any similar method is heavily discouraged.

fxtressia: > Note that `destroy` is not a thing in gtk4 (except with windows).
> 
> What are you trying to acomplish here? Remove a row from a listbox? In such a case there is gtk_list_box_remove (or gtk_box_remove).

Okayy

fxtressia: > Note that `destroy` is not a thing in gtk4 (except with windows).
> 
> What are you trying to acomplish here? Remove a row from a listbox? In such a case there is gtk_list_box_remove (or gtk_box_remove).

When I wrote that code, I didn't properly read the docs so I just... add the if sentence. ^^

fxtressia: Sorry!!

fxtressia:
> > @A6GibKm Also, I'm getting this issue where Flatseal actually doesn't use the `styles.css` file. Could you help investigate? Thanks.
> 
> This bit here
> 
> ```js
> var FlatsealApplication = GObject.registerClass({
>     GTypeName: 'FlatsealApplication',
> }, class FlatsealApplication extends Gtk.Application {
>     _init() {
>         super._init({
>             application_id: 'com.github.tchx84.Flatseal',
>             flags: Gio.ApplicationFlags.FLAGS_NONE,
>         });
> 
>         this._window = null;
> 
>     }
> ```
> 
> it should extend Adw.Application this will auto load RESOURCE_BASE_PATH/style.css, so bellow the `flags` line add `resource_base_path: '/com/github/tchx84/Flatseal/',`.

Oops!! Did I not subclass it from Adw.Application? Sorry 😭

fxtressia:
@A6GibKm I think the &amp became & in Todoist: To-Do List & Task

So we need to replace it with &amp again I think?

fxtressia:
> The window has a minimal width of 364px, this won't work in mobile, it should reach 360px.

Umm.. How do I fix this? I found no references of `364` in VSCode except in the svg logo.

Maybe it's implicitly declared?

A6GibKm:
> @A6GibKm I think the &amp became & in Todoist: To-Do List & Task
> 
> So we need to replace it with &amp again I think?

Yes. Normally when you parse xml/yaml the output is unscaped so one has to escape it again.

Glib has a function to escape strings for this purpose. But I couldnt find which label has use-markup active either after a quick glance.

A6GibKm:
> > The window has a minimal width of 364px, this won't work in mobile, it should reach 360px.
> 
> Umm.. How do I fix this? I found no references of `364` in VSCode except in the svg logo.
> 
> Maybe it's implicitly declared?

What happens is that that there is at some point a widget that requires a minimal width that cannot be satisfied otherwise, this widget could be in either side of the leaflet so basically anywhere in the app. I will have a look at this issue later so don't worry too much about it for now.

For exaple two linked horiontal boxes with a minimal width of 180 each, with margins of 2px so you are not even going to see the explicit 364.

A6GibKm:
> > > The window has a minimal width of 364px, this won't work in mobile, it should reach 360px.
> > 
> > 
> > Umm.. How do I fix this? I found no references of `364` in VSCode except in the svg logo.
> > Maybe it's implicitly declared?
> 
> What happens is that that there is at some point a widget that requires a minimal width that cannot be satisfied otherwise, this widget could be in either side of the leaflet so basically anywhere in the app. I will have a look at this issue later so don't worry too much about it for now.
> 
> For exaple two linked horiontal boxes with a minimal width of 180 each, with margins of 2px so you are not even going to see the explicit 364.

By the way the problem fixed itself after the last batch of commits.

A6GibKm:
@tchx84 Hi, is there anything I can help with to get this PR to the finish line?

I can start by fixing the remaining issues and rebasing. Is there anything missing on your side?

fxtressia:
> @tchx84 Hi, is there anything I can help with to get this PR to the finish line?
> 
> I can start by fixing the remaining issues and rebasing. Is there anything missing on your side?

Feel free to work on this!

tchx84:
@A6GibKm please go ahead !

A6GibKm:
mmm the "Other files" is indeed borked, ill look into it later.

tchx84:
This work is being continued here https://github.com/tchx84/Flatseal/pull/515 , so I am closing this.

Thanks a lot @natasria for working on this.

A6GibKm:
Thanks @natasria !

PR #473, opened 2022-09-23
manifest: Bump to GNOME 43

tchx84:


SuperNinja-4965:
You are probably aware of this but after building the manifest with the version bump (at least for me) when i go to run flatseal it throws an error about not being able to find webkit2 and closes immediately.

tchx84:
@joshuag1000 I did :sweat_smile:  but ran out of time last night, thanks for jumping in though.

I fixed it now, would greatly appreciate if you could help me with further testing :raised_hands:

SuperNinja-4965:
I'll happily help with testing.
seems like i missed the obvious when i went to try to fix it myself! lol.

tchx84:
@joshuag1000 trust me it wasn't obvious, lol..

SuperNinja-4965:
I've just noticed one or two small things but i am unsure if they are related to this update.
- The show details button does not work.
- The app doesn't appear to read overrides set in /var/lib/flatpak/overrides
- i'm unable to tell if the portal settings are working. But i could just be looking in the wrong place here.
- The reset button is always active regardless of whether i have set overrides and it appears to just turn off the run in background toggle which appears to be on for every app. (this one could have just been something i did but i'm not so sure).

tchx84:
> I've just noticed one or two small things but i am unsure if they are related to this update.

I suspect most of the issues aren't related to the runtime bump, but, we can confirm by testing with the current flathub version.

> 
> * The show details button does not work.

Ill investigate, this seems to happen with the current flathub version as well.

> * The app doesn't appear to read overrides set in /var/lib/flatpak/overrides

This is expected.

> * i'm unable to tell if the portal settings are working. But i could just be looking in the wrong place here.

Yeah, there is isn't a straightforward way to test a particular portal, because it kind of depends on each individual app behavior as well, but Flatseal-wise, basically if the changes you make are restored properly afterwards should be enough.

tchx84:
Ill merge this as-is, and follow-up with those issues separately.

SuperNinja-4965:
> Ill merge this as-is, and follow-up with those issues separately.

Makes sense. I haven't found anything else other than what i have mentioned and like you said those most likely aren't caused by the version bump.

PR #489, opened 2022-10-04
po: update Chinese (China) translation

EricZhang456:


EricZhang456:
I didn't see #480. nvm

PR #498, opened 2022-10-28
Add fixed Finnish translation

karpse:
Add fixed Finnish translation.


tchx84 (changes requested):
Looks good in general, but I still see some non-translated strings? Isn't there translations for things like "changed by the user"? I guessed there should be since that's a very little technical term.

PR #503, opened 2022-12-30
Add Belarusian translation

BraveFern:


tchx84 (changes requested):
Hey @vokaEntropy , thanks for working on this, by any chance do you why I am getting this message?

```
$ LC_ALL=be_BY.UTF-8 flatpak --user run --branch=master com.github.tchx84.Flatseal

(com.github.tchx84.Flatseal:2): Gtk-WARNING **: 21:00:00.963: Whoever translated default:LTR did so wrongly. Defaulting to LTR.
```

po/be.po
```diff
@@ -0,0 +1,581 @@
...
+msgid "Portal Version"
+msgstr "Версія партала"
+
+#: src/widgets/globalInfoViewer.ui:183
+msgid "Changes Path"
+msgstr "Шлях змен"
+
+#: src/widgets/menu.ui:7
+msgid "_Help"
+msgstr "_Даведка"
```
tchx84, line 351: You need to "mark" different characters for the shortcuts, otherwise you can access the second one (it becomes ambiguous)

image

tchx84 (changes requested):

PR #506, opened 2023-01-30
Added information for RHEL 9 compability 

rizajur:
updated README.md

preview : https://github.com/rizajur/Flatseal/blob/master/README.md

 for dirty workaround regarding broken font implementation of OS default interface font (Gnome 40.1)
would be cool if we could change the font family to something dejavu related in the future depending on the availability of the required packages in the different operating systems or if embedding is best option

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,8 +12,15 @@ For more details please visit the [documentation](./DOCUMENTATION.md) page.
 
 ### RHEL 9 Compatiblity [Almalinux and RockyLinux, etc]
 In case the application font texts are not displayed correctly please use the
+
+'gsettings set org.gnome.desktop.interface font-name "DejaVu Sans Regular 10"'  where the 10 is the font size
+  
+or
+
 `gnome-tweaks` package to **change the font for Interface Text from the default to
 something silimar like** `dejavu sans regular`.
+
+
 
 
 ## Get it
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,6 +10,10 @@ Simply launch Flatseal, select an application and modify its permissions. Restar
 
 For more details please visit the [documentation](./DOCUMENTATION.md) page.
 
+### RHEL 9 Compatiblity [Almalinux and RockyLinux, etc]
+In case the application font texts are not displayed correctly please use the `gnome-tweaks` package to **change the font for Interface Text from the default to something silumar like** `dejavu sans regular`
```
rusty-snake:
```suggestion
In case the application font texts are not displayed correctly please use the
`gnome-tweaks` package to **change the font for Interface Text from the default to
something silimar like** `dejavu sans regular`.
```

Break long line and fix typo

rusty-snake:
```bash
gsettings set org.gnome.desktop.interface font-name "<FONT>"
```

Should work w/o installing additional software. (untested)

tchx84:
Apologies for the late review, would you mind moving this to the documentation file instead? Under the "tips and tricks" section, or add a "Troubleshoot" section.

tchx84 (changes requested):
See comment above.

PR #508, opened 2023-03-01
bin: workaround to enable the use of promises

nyx-lyb3ra:
This small workaround (documented at https://gitlab.gnome.org/GNOME/gjs/-/issues/468) enables us to use promises in the application.

This will no longer be necessary with the GNOME 44 SDK, since there we will be able to use `Gio.Application.prototype.runAsync()`.


tchx84:
I just bumped the runtime to GNOME 44, so I guess this is no longer necessary.

PR #515, opened 2023-03-31
Gtk4 port part 2

A6GibKm:
This is a continuation of https://github.com/tchx84/Flatseal/pull/464

Missing:

- [x] FlatsealPermissionEntryRow: Needs to be a expander row with EntryRows as children (implemented it in another fashion)
- [ ] DocView: The background needs a new color
- [x] ShowDetails opens software twice for some reason (Pre-existing issue)
- [x] Markup error somewhere
- [x] Port notification to AdwToast
- [x] Entry Completions are probably borked
- [x] Close buttons on rows inherit the accent_color, this is wrong


tchx84:
@A6GibKm awesome, I'll give it a try today and come back.

A6GibKm:
> DocView: The background needs a new color

Where would one set this? I see there is a stylesheet in index.html, but it seems highly non-editable.

tchx84:
> > DocView: The background needs a new color
> 
> Where would one set this? I see there is a stylesheet in index.html, but it seems highly non-editable.

This is generated from Apostrophe's export MD to HTML option, plus a hack I added to remove the embedded fonts. I basically run this https://gitlab.gnome.org/tchx84/apostrophe/-/tree/flatseal .

If it's too troublesome we can leave that for later.

A6GibKm:
Might I ask what code reveals the  in-app notification? I cannot find it.

tchx84:
> Might I ask what code reveals the in-app notification? I cannot find it.

It's the widget itself connecting to the permissions model signals, see https://github.com/tchx84/Flatseal/blob/master/src/widgets/undoPopup.js#L42

tchx84 (changes requested):
It looks quite good, here's a quick dump of a few details I noticed:

* keyboard: Navigation gets stuck or certain UI elements can't be focused, e.g, use arrows or tab to navigate trough the UI and it will get stuck on apps list or permissions list.
* keyboard: Activation of the main menu (F10) doesn't work e.g, pressing F10 won't open the menu.
* keyboard: Menu items mnemonics don't show up e.g, pressing ALT while menu it's open, won't make the mnemonics show up.
* keyboard: Activation of apps search (CTRL+F) doesn't work e.g, only seems to work if the search button is focused.
* window: When in "compact" mode, can't swipe back from the permissions view to the apps view.
* window: The lower "show details and reset" menu would stay around after the window is back at full size, e.g. switch the window to compact mode, go to the permissions view and then double click the header bar to maximize back to full size.
* permissions: The "other files" section doesn't seem to reset properly e.g, select Flatsealon the apps list, check the "other files" section, now re-select Flatseal multiple times on the apps list, and a lot of duplicate entries will appear on the "other files" section.
* permissions: In addition to the issue mentioned above, these duplicates seem to persist, switching to another app and back doesn't seem to fix it. This doesn't seem to happen at the model level (reset button is kept inactive), so not entirely sure what's going on.
* permissions: In addition to the above, the issue seems to happen to other similar widgets e.g.  dbus, variables, persist, etc.
* documentation window: Window title shows app id not the app name.
* documentation window: keyboard shortcuts don't seem to work e.g, CTRL+F won't activate search entry, CTRL+Q won't close the window.

src/widgets/aboutDialog.js
```diff
@@ -19,27 +19,11 @@
  */
 
 const {GObject, Gtk} = imports.gi;
+const ABOUT_RESOURCE = "/com/github/tchx84/Flatseal/widgets/aboutDialog.ui";
+function showAboutDialog(window) {
+    let builder = Gtk.Builder.new_from_resource(ABOUT_RESOURCE);
```
DaPigGuy: Code styling nit, but these should be declared with the `const` keyword

A6GibKm: Thanks, its not my code, but ill fix it anyways. I have no clue about gjs or js so I appreciate these style/good practices comments. Fixed.

A6GibKm:
> keyboard: Navigation gets stuck or certain UI elements can't be focused, e.g, use arrows or tab to navigate trough the UI and it will get stuck on apps list or permissions list.

Could you be more precise on which elements please? Here it works, but there are a few GTK 4 rough edges, like going from the listbox on the left to the header bar.  Or going from left panel to right panel using TAB only.

You have no clue how helpful this review was. I found multiple things that were broken in more apps and simple fixes for them.

> keyboard: Activation of the main menu (F10) doesn't work e.g, pressing F10 won't open the menu.

fixed, was it working in gtk3? if so how?

> keyboard: Menu items mnemonics don't show up e.g, pressing ALT while menu it's open, won't make the mnemonics show up.

GTK 4 quirk, they appear when starting keynav inside of them, ill try to re-open upstream the issue as it is indeed weird.

> window: When in "compact" mode, can't swipe back from the permissions view to the apps view.

What I can reproduce here, is that swiping back goes to the separator (which looks like a grey page) instead of the list of apps, I fixed this. Could you confirm?

> permissions: The "other files" section doesn't seem to reset properly e.g, select Flatseal on the apps list, check the "other files" section, now re-select Flatseal multiple times on the apps list, and a lot of duplicate entries will appear on the "other files" section.

This and related items should be fixed in a later commit, try fetching latest commit and please confirm if the issue was indeed fixed.

> documentation window: Window title shows app id not the app name.

Fixed, incidentally, the app name was the app id, I fixed this too.

> documentation window: keyboard shortcuts don't seem to work e.g, CTRL+F won't activate search entry, CTRL+Q won't close the window.

Fixed

EDIT:

> window: The lower "show details and reset" menu would stay around after the window is back at full size, e.g. switch the window to compact mode, go to the permissions view and then double click the header bar to maximize back to full size.

Fixed

A6GibKm:
Also CI is broken, no idea how to fix this one as it uses an image. Probably we can use https://github.com/bilelmoussaoui/ashpd/tree/master/.github/workflows.

EDIT: Used just that and CI passes. I have yet to try the bundle.

A6GibKm:
I found the following thing, the about dialog has a list of translators, what we usually do is to add a 
```
    <property name="translator_credits" translatable="yes" comments="Add your name to the translator credits list">translator-credits</property>
```
each translation adds the credit for the specific translators in use instead of having a list of all translations. Do you want to implement this change in this PR?

A6GibKm:
Only things missing from my side are 

- Using `#242424` (dark mode) and `#fafafa` (light mode), which correspond to `@window_bg_color`, in the document viewer. This is out of scope for the PR. Probably also check @accent_color for links. It might be ideal if the css file is in its own file so its easier to edit.
- Checking if there are further key navigation issue that are not intrinsic GTK 4 issues
- Libadwaita allows setting accent colors different from blue, are you interested in doing so?

See https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/named-colors.html

.github/workflows/CI.yml
```diff
@@ -4,9 +4,17 @@ on:
...
     steps:
-    - uses: actions/checkout@master
-    - uses: tchx84/gtk-apps-testing-docker-action@gjs
+    - uses: actions/checkout@v3
+    - uses: flatpak/flatpak-github-actions/flatpak-builder@v5
+      with:
+        bundle: "flatseal.flatpak"
+        manifest-path: "com.github.tchx84.Flatseal.json"
+        run-tests: "true"
+        cache-key: flatpak-builder-${{ github.sha }}
```
tchx84: The tests are "passing"  because jasmine-gjs is not included here, and therefore the actual tests are not running. 

I am working on this here https://github.com/tchx84/Flatseal/pull/514 it has it owns issues. For this PR we should aim to have the tests pass with the current CI setup.

You should be able to run this with a fedora toolbox, installing these dependencies https://github.com/tchx84/gtk-apps-testing-docker-action/blob/gjs/Dockerfile and doing this https://github.com/tchx84/gtk-apps-testing-docker-action/blob/gjs/entrypoint.sh

src/widgets/aboutDialog.js
```diff
@@ -19,27 +19,11 @@
  */
 
 const {GObject, Gtk} = imports.gi;
+const ABOUT_RESOURCE = "/com/github/tchx84/Flatseal/widgets/aboutDialog.ui";
+function showAboutDialog(window) {
+    const builder = Gtk.Builder.new_from_resource(ABOUT_RESOURCE);
 
-
-var FlatsealAboutDialog = GObject.registerClass({
```
tchx84, line 24: Just curious, why not simply sub classing AdwAboutDialog here?

src/widgets/window.js
```diff
@@ -96,47 +95,50 @@ var FlatsealWindow = GObject.registerClass({
...
         this._detailsActionButton = new FlatsealDetailsButton(this._permissions);
-        this._startActionBox.add(this._detailsActionButton);
+        this._startActionBox.append(this._detailsActionButton);
         const resetActionButton = new FlatsealResetButton(this._permissions);
-        this._endActionBox.add(resetActionButton);
+        this._endActionBox.append(resetActionButton);
 
-        this._undoPopup = new FlatsealUndoPopup(this._permissions);
-        this._undoPopupBox.add(this._undoPopup);
+        this._toast = new Adw.Toast();
```
tchx84, line 102: can't we sub class Adw.Toast and keep these things there?

tchx84:
@A6GibKm thanks for the great work, responding on questions here:

> Could you be more precise on which elements please? Here it works, but there are a few GTK 4 rough edges, like going from the listbox on the left to the header bar. Or going from left panel to right panel using TAB only.

With GTK 3, we can move around all the main windows elements by using TAB or the keyboard arrows. With GTk 4, even with the TAB quirk, I wasn't able to go back to elements like the search box or the main menu.

> fixed, was it working in gtk3? if so how?

An accelerator on the window.ui file.

> GTK 4 quirk, they appear when starting keynav inside of them, ill try to re-open upstream the issue as it is indeed weird.

Thanks.

> What I can reproduce here, is that swiping back goes to the separator (which looks like a grey page) instead of the list of apps, I fixed this. Could you confirm?

hmm, in my case the panel wasn't moving at all, I would swipe and it would stay on the permissions panel.

> This and related items should be fixed in a later commit, try fetching latest commit and please confirm if the issue was indeed fixed.

Sure, I'll try ASAP.

> each translation adds the credit for the specific translators in use instead of having a list of all translations. Do you want to implement this change in this PR?

 Not entirely sure I understand how this would work, is that something that would make sense if translations are handled externally (e.g damnlies)? I prefer to mention ALL people who contributed to Flatseal, it's an important part of how this project is driven.
 
 > Libadwaita allows setting accent colors different from blue, are you interested in doing so?
 
 We can do this in a separate PR.

src/widgets/aboutDialog.js
A6GibKm: AboutWindow and AboutDialog are final classes, they cannot be subclassed

src/widgets/window.js
A6GibKm: no, its final. but I feel it is overkill if we only need to set three lines of properties

A6GibKm:
> Not entirely sure I understand how this would work, is that something that would make sense if translations are handled externally (e.g damnlies)? I prefer to mention ALL people who contributed to Flatseal, it's an important part of how this project is driven.

There is a translatable string called `_("translators-credits")` used in the about dialog, which is special in gettext. The idea is that every time you work on a translation you add your name to this translatable string and this the dialog will show all people that contributed to your language translation. 

I fear that doing this will prove difficult as it will imply adding the names to each pot file individually to not lose the ones that are already there.

src/widgets/docsViewer.js
```diff
@@ -109,14 +109,25 @@ var FlatsealDocsViewer = GObject.registerClass({
...
     }
 
+    _OpenUriCb(launcher, res) {
+        try {
+            launcher.launch_finish(res);
+        } catch(err) {
+            logError(err);
+        }
+    }
+
```
tchx84, line 130: To make this PR simpler to review and merge, I suggest to leave misc improvements out. Let's just focus strictly on migrating to GTK 4 and Libadwaita.

.github/workflows/CI.yml
A6GibKm: I am trying but no idea how it works. It might be easier to just update the container in use for now

tchx84:
@A6GibKm I re-tested everything again, here's the summary

### From previous reviews:

> keyboard: Navigation gets stuck or certain UI elements can't be focused, e.g, use arrows or tab to navigate trough the UI and it will get stuck on apps list or permissions list.

Fixed

> keyboard: Activation of the main menu (F10) doesn't work e.g, pressing F10 won't open the menu.

Fixed 

> keyboard: Menu items mnemonics don't show up e.g, pressing ALT while menu it's open, won't make the mnemonics show up.

Reported to upstream.

> keyboard: Activation of apps search (CTRL+F) doesn't work e.g, only seems to work if the search button is focused.

Fixed

> window: When in "compact" mode, can't swipe back from the permissions view to the apps view.

**Still not working**

Do I need to enable something special to swipe with touchpad on libadwaita?

> window: The lower "show details and reset" menu would stay around after the window is back at full size, e.g. switch the window to compact mode, go to the permissions view and then double click the header bar to maximize back to full size.

Fixed

> permissions: The "other files" section doesn't seem to reset properly e.g, select Flatsealon the apps list, check the "other files" section, now re-select Flatseal multiple times on the apps list, and a lot of duplicate entries will appear on the "other files" section.

Fixed

> permissions: In addition to the issue mentioned above, these duplicates seem to persist, switching to another app and back doesn't seem to fix it. This doesn't seem to happen at the model level (reset button is kept inactive), so not entirely sure what's going on.

Fixed

> permissions: In addition to the above, the issue seems to happen to other similar widgets e.g. dbus, variables, persist, etc.

Fixed

> documentation window: Window title shows app id not the app name.

Fixed

> documentation window: keyboard shortcuts don't seem to work e.g, CTRL+F won't activate search entry, CTRL+Q won't close the window.

**Still not working**

### New / Still happening

* window: Can't close the window with CTRL+Q
* documentation window: Can't close the window with CTRL+Q
* permissions: "Overridden by" badges are positioned on the wrong row, e.g. go to Flatseal, "other files" add a new row "/newpath", then you will see:

Screenshot from 2023-04-03 10-29-20

.github/workflows/CI.yml
tchx84: Just to make sure there's no misunderstanding here: 

All this change is doing is simply disabling the tests. That's not a solution.

That GH action is not working properly either, reason why I am investigating it on https://github.com/tchx84/Flatseal/pull/514 , so let's separate that issue from this PR. This PR is big enough as it is. So,

**Please remove this change**, or we're are simply going to get stuck here. 

I suggest you try:

1. Creating a Fedora toolbox
2. Installing the dependencies as per https://github.com/tchx84/gtk-apps-testing-docker-action/blob/gjs/Dockerfile (copy paste)
3. Running "ninja test"

If there's something that is not working or clear, just ask specific questions.

tchx84:
One more thing I noticed, if you mimic the scenario where, for whatever reason, Flatseal can't find any applications... the UI ends up in a inconsistent state. To reproduce:

1. Open Flatseal
2. Select Flatseal from the applications list
3. Go to "other files"
4. Remove all file system permissions
5. Restart Flatseal

You will notice the app ends up a in strange state, e.g, the reset button is active although there's no application selected.

Also, a few errors are showing up:

```
(com.github.tchx84.Flatseal:2): Gjs-CRITICAL **: 11:10:30.303: JS ERROR: TypeError: this._window._shutdown is not a function
_quit@resource:///com/github/tchx84/Flatseal/js/application.js:71:22
main@resource:///com/github/tchx84/Flatseal/js/main.js:37:24
run@resource:///org/gnome/gjs/modules/script/package.js:206:19
```

AND

```
(com.github.tchx84.Flatseal:2): Gjs-CRITICAL **: 11:10:44.346: JS ERROR: Error: Expected type filename for Array element but got type 'undefined'
getMetadataPathForAppId@resource:///com/github/tchx84/Flatseal/js/models/applications.js:332:21
_loadPermissions@resource:///com/github/tchx84/Flatseal/js/models/permissions.js:182:32
_setup@resource:///com/github/tchx84/Flatseal/js/models/permissions.js:274:14
reset@resource:///com/github/tchx84/Flatseal/js/models/permissions.js:326:14
_clicked@resource:///com/github/tchx84/Flatseal/js/widgets/resetButton.js:46:27
main@resource:///com/github/tchx84/Flatseal/js/main.js:37:24
run@resource:///org/gnome/gjs/modules/script/package.js:206:19
Caused by: Error: Invalid element in array
```

A6GibKm:
>   window: When in "compact" mode, can't swipe back from the permissions view to the apps view.

It turns out it is a GTK 4 issue with scrolled windows, it seems its rather involved.

> Reported to upstream.

please share link

> documentation window: keyboard shortcuts don't seem to work e.g, CTRL+F won't activate search entry, CTRL+Q won't close the window.

Fixed, it was a mistake on my end by removing the `_shutdown` fn in window.js. is it expected that `_shutdown` fns can be called from outside their scope? In python `_` means that the fn is private to the class (but nothing enforces this).

> One more thing I noticed, if you mimic the scenario where, for whatever reason, Flatseal can't find any applications... the UI ends up in a inconsistent state. To reproduce:

ugh I cannot reproduce this

> Also, a few errors are showing up:

This was part of ctrl+q not working, fixed

> Also, a few errors are showing up:

mmm I havent seen it, how can this be reproduced?

A6GibKm:
> permissions: "Overridden by" badges are positioned on the wrong row, e.g. go to Flatseal, "other files" add a new row "/newpath", then you will see:

Cannot reproduce.

src/widgets/pathsViewer.js
```diff
@@ -132,8 +125,8 @@ var FlatsealPathsViewer = GObject.registerClass({
         row.text = path;
         row.connect('remove-requested', this._remove.bind(this, row));
         row.connect('notify::text', this._changed.bind(this));
-        row.show();
 
-        this._box.pack_end(row, true, true, 0);
+        this.append(row);
```
DaPigGuy:
```suggestion
        this.prepend(row);
```

Fixes the issue:
> permissions: "Overridden by" badges are positioned on the wrong row, e.g. go to Flatseal, "other files" add a new row "/newpath", then you will see:

---

Changing from GTK3's GtkBox.pack_end to GTK4's GtkBox.append resulted in the ordering of the child rows being reversed. The existing logic to set the status icon for the rows is reliant on said ordering, hence the incorrectly set icons (on opening an app's page with existing overrides). This is the easiest fix and maintains the existing UI :slightly_smiling_face:

src/widgets/permissionEntryRow.js
```diff
@@ -31,19 +31,19 @@ var FlatsealPermissionEntryRow = GObject.registerClass({
         'permission',
         'box',
         'button',
-        'image',
     ],
-}, class FlatsealpermissionEntryRow extends Handy.PreferencesRow {
+}, class FlatsealpermissionEntryRow extends Adw.PreferencesRow {
```
DaPigGuy, line 35:
```suggestion
}, class FlatsealPermissionEntryRow extends Adw.PreferencesRow {
```

Noticed while debugging the issue :sweat_smile:

src/widgets/pathsViewer.js
A6GibKm: The problem was about the badge, not the row. In any case its kinda weird UI to prepend rows, ill fix it to match previous behaviour

src/widgets/permissionEntryRow.js
A6GibKm: Whats affected by this? Its a preexisting issue and I fear breaking something by changing this

src/widgets/pathsViewer.js
DaPigGuy: https://github.com/A6GibKm/Flatseal/blob/gtk4-port/src/widgets/pathsViewer.js#L90 Yes, I'm aware, the logic for setting the badge was dependent on the ordering of the rows

src/widgets/permissionEntryRow.js
DaPigGuy: Nothing should be affected by this, there are no references to this with that exact casing, and the class is simply just passed to GObject.registerClass

tchx84:
@A6GibKm great work, I did another pass and these are the remaining issues with the PR:

* ci: Changes have not been reverted and tests are still not passing. Please read https://github.com/tchx84/Flatseal/pull/515#discussion_r1156068922 and let me know if for whatever reason you can't work on this.
* window: The reset button is active when Flatseal launches without any filesystem permissions, when it shouldn't be. To reproduce this: use flatpak-builder and flatpak-run with this branch, remove all filesystem permissions from Flatseal and re-start Flatseal.
* permissions: The "Overridden by" badge no longer use a different colors to differentiate between "global" and "user" overrides. To reproduce this: Add any global override, select any other application and then you will notice the badge color is the same as with changes made by the user.
* documentation window: CTRL+Q closes the whole application, not only the documentation window.

Let's not add any other non essential changes from now on, so we can start stabilizing this branch.

A6GibKm:
> ci: Changes have not been reverted and tests are still not passing. Please read https://github.com/tchx84/Flatseal/pull/515#discussion_r1156068922 and let me know if for whatever reason you can't work on this.

I can do it locally, but would that be enough on your end? I would prefer to avoid dealing with CI

> documentation window: CTRL+Q closes the whole application, not only the documentation window.

This might be a different behaviour from gtk3, but in fact its the expected behavior of the shortcut, as it is meant to kill the app. EDIT: It is also what other apps try to do when they do not forget to set the `application` property of the window. 

Do note that this comes from `this.set_accels_for_action('app.quit', ['<Control>q']);`, `app.` actions are app wide

tchx84:
@A6GibKm 

> I can do it locally, but would that be enough on your end? I would prefer to avoid dealing with CI

Yeah, that's no problem, just give me access to this branch and we can work together on that.

A6GibKm:
> @A6GibKm
> 
> > I can do it locally, but would that be enough on your end? I would prefer to avoid dealing with CI
> 
> Yeah, that's no problem, just give me access to this branch and we can work together on that.

I think I invited you to my repo, no idea if thats enough. This is one instance where gitlab >> github.

A6GibKm:
> permissions: The "Overridden by" badge no longer use a different colors to differentiate between "global" and "user" overrides. To reproduce this: Add any global override, select any other application and then you will notice the badge color is the same as with changes made by the user.

I cannot see global overrides for some reason but I think I fixed this.

A6GibKm:
> window: The reset button is active when Flatseal launches without any filesystem permissions, when it shouldn't be. To reproduce this: use flatpak-builder and flatpak-run with this branch, remove all filesystem permissions from Flatseal and re-start Flatseal.

Ugh, I was able to reproduce it once, but not anymore, maybe the fix to the rows going from append to prepend helped here? Well I also found a possible issue that might have fixed it

DaPigGuy:
> I think I invited you to my repo, no idea if thats enough. This is one instance where gitlab >> github.

Just so you're aware, this is an option at the bottom of the right sidebar:
image

A6GibKm:
~~The test is getting timeout, no idea why. At least it runs locally.~~

EDIT: Test are working locally.

tchx84:
@A6GibKm I fixed CI bits and tests are passing now, excellent work!

Let's "freeze" this PR now so I can start doing some cleanups before we merge it.

A6GibKm:
> @A6GibKm I fixed CI bits and tests are passing now, excellent work!
> 
> Let's "freeze" this PR now so I can start doing some cleanups before we merge it.

Ok! Last changes were me trying to remove some unneeded diffs from the earlier commits, but rebasing this PR is not really easy. Besides wiring up a new CI and cleaning the history this should be ready.

EDIT: I see CI passing, did you update the image?

tchx84 (approved):

tchx84:
@A6GibKm great work :partying_face:

tchx84:
> > @A6GibKm I fixed CI bits and tests are passing now, excellent work!
> > Let's "freeze" this PR now so I can start doing some cleanups before we merge it.
> 
> Ok! Last changes were me trying to remove some unneeded diffs from the earlier commits, but rebasing this PR is not really easy. Besides wiring up a new CI and cleaning the history this should be ready.
> 
> EDIT: I see CI passing, did you update the image?

I did :)

PR #542, opened 2023-05-01
style: Fix applications rows padding

### LINKED ISSUE #537: Sidebar item padding
The sidebar list items have too much horizontal, and not enough vertical padding. I'd reduce the horizontal padding by 12px, and add 3 or 6px vertically.

image

tchx84:
Closes #537

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,3 +1,7 @@
+.navigation-sidebar row {
+    padding: 3px 0px;
+}
+
 .permissions clamp {
     margin-bottom: 25px;
 }


tchx84:
@bertob strangely, the current styling is what we get from simply applying `navigation-sidebar` class to the list box. The other detail is that we're using AdwActionRows (which is what we needed to tweak here).

Before 
Screenshot from 2023-05-01 10-17-22

After
Screenshot from 2023-05-01 10-17-47

what you think?

bertob:
Definitely an improvement, but the margin around the boxes looks too narrow (looks like 3px, should be 6px).

tchx84:
OK, I think it's now closer to what gnome settings looks like (but without adding excessive height per row).

After

Screenshot from 2023-05-02 10-08-59

(btw, I think there's a bug in GNOME 44 screenshot tool, it's cropping a few pixels from the left side)

PR #570, opened 2023-09-13
Update Ukrainian localization

ghost:


tchx84 (changes requested):
Thanks @Vovkiv , it seems you updated an older version of this file... can you apply your changes to the latest version?

ghost:
Now I think everything should be alright?
I just fixed newly added strings + some unrelated changes.

Or wait, I think I can quickly fix.

ghost:
Now, I think, everything should be correct.

tchx84 (approved):
@Vovkiv thanks, working now !

PR #571, opened 2023-09-13
Corrected Dutch translation

Vistaus:
Could you please leave the new strings blank next time? I don't like it when developers that don't speak the language automatically translate them. As you know from previous releases, I always update the translation in time.


tchx84:
surely, thanks for fixes!

PR #587, opened 2023-09-22
Update German translation

daPhipz:
I also added some No-Break space characters where sensible.


tchx84 (approved):
@daPhipz excellent, thanks for the extra fixes

PR #614, opened 2023-10-30
Add "override" keyword to desktop file

wjt:
Flatseal is a fancy interface for `flatpak override` but the word
“override” did not previously appear in its metadata.


tchx84 (approved):

tchx84:
@wjt indeed! thanks

PR #615, opened 2023-11-06
Fix french translation of network

LSerranoPEReN:
The previous translation of "Network" in the French translation file is « Réseau sans fil », which meant "Wireless network". This commit update the translation to a more accurate one.


tchx84:
Thanks!

PR #617, opened 2023-11-26
models: update for AppStream 1.0

yselkowitz:
AppStream 1.0.0 includes several API breaks compared to the previous 0.16 (and earlier) versions, but unfortunately both versions used the same 1.0 GI version, so there is no way to differentiate at import time. ~~As such, try/catch is used to fall back to the 0.16 API where they differ.~~ Therefore, enforce the requirement during the build to make this clear.

### DIFF
--- a/src/models/applications.js
+++ b/src/models/applications.js
@@ -326,14 +326,18 @@ var FlatpakApplicationsModel = GObject.registerClass({
         if (component.get_name())
             appdata.name = component.get_name();
 
-        if (component.get_developer_name())
-            appdata.author = component.get_developer_name();
+        const developer = component.get_developer();
+        if (developer && developer.get_name())
+            appdata.author = developer.get_name();
 
         const launchable = component.get_launchable(AppStream.LaunchableKind.DESKTOP_ID);
         if (launchable && launchable.get_entries())
             [appdata.launchable] = launchable.get_entries();
 
-        const [release] = component.get_releases();
+        const releaselist = component.get_releases_plain();
+        if (!releaselist)
+            return appdata;
+        const [release] = releaselist.get_entries();
         if (!release)
             return appdata;
 
--- a/src/meson.build
+++ b/src/meson.build
@@ -1,6 +1,7 @@
 pkgdatadir = join_paths(get_option('datadir'), meson.project_name())
 gnome = import('gnome')
 
+dependency('appstream', version : '>= 1.0')
 dependency('libadwaita-1', version : '>= 1.5')
 dependency('webkitgtk-6.0', version : '>= 2.40')
 


src/models/applications.js
```diff
@@ -319,14 +319,32 @@ var FlatpakApplicationsModel = GObject.registerClass({
...
+        try {
+            /* AppStream 1.0.0+ */
+            const releaselist = component.get_releases_plain();
+            if (!releaselist)
+                return appdata;
+            [release] = releaselist.get_entries();
+        } catch (err) {
+            /* AppStream up to 0.16 */
+            [release] = component.get_releases();
+        }
```
tchx84: Thanks for working on this, can you please modify this to assume it's running with AppStream 1.0.0+  only? (no try-except block, no back-compatibility).

I will merge it once the gnome runtime bumps AppStream to that version and, in the meantime, you should contact whoever is distributing flatseal outside of flathub to patch this downstream.

Version guessing is not something I want upstream.

tchx84 (changes requested):

darkblaze69:
with this patch I still get an error with failing launch

`JS ERROR: TypeError: component.get_developer_name is not a function`

tchx84 (approved):
This looks much better, thanks. I will merge this as soon as GNOME runtime upgrades too.

C0rn3j:
@tchx84 is this still waiting for a runtime upgrade?

Just had to include this in [AUR/flatseal](https://aur.archlinux.org/packages/flatseal) because of the crash on `component.get_developer()`.

tchx84:
> @tchx84 is this still waiting for a runtime upgrade?
> 
> Just had to include this in [AUR/flatseal](https://aur.archlinux.org/packages/flatseal) because of the crash on `component.get_developer()`.

Yes, this change is still waiting for the runtime upgrade. Be assured that this will be merged once it happens. Please let me know if notice anything else.

tchx84:
Unfortunately, GNOME 46 still ships with older appstream version.

```
$ flatpak run --command=appstreamcli org.gnome.Sdk//46 --version
AppStream version: 0.16.4
```

C0rn3j:
Forgive if I am misunderstanding something here.

Wouldn't it be a better idea to pull this, making the places that do not support the current stable API run into issues, making them figure out there is a new stable API and having to apply workarounds?

Currently it's exactly the other way around, which seems to be punishing those who ship the current stable release.

C0rn3j:
This now conflicts after the 2.2.0 release.

EDIT: It's just a minor 1.4 -> 1.5 string change

TheComputerGuy96:
GNOME 47 runtime has just been released: https://gitlab.gnome.org/GNOME/gnome-build-meta/-/tags/47.0 (so this MR can finally move forward)

~~Edit: Well except for the CI which requires https://github.com/andyholmes/flatter/pull/111~~ That MR got merged

tchx84:
@yselkowitz thanks for the patience , merged with https://github.com/tchx84/Flatseal/pull/710

PR #633, opened 2024-01-17
screenshots: Update

### LINKED ISSUE #612: Screenshots too similar
Screenshots https://github.com/tchx84/Flatseal/blob/master/screenshots/en/1.png and https://github.com/tchx84/Flatseal/blob/master/screenshots/en/2.png might be too similar and confuse users due to that?

### LINKED ISSUE #622: Better screenshots
Appdata screenshots should not be fullscreen (use the "window screenshot" option in GNOME Shell), and 1000x700 px or smaller. For more details see the new [Flathub guidelines](https://docs.flathub.org/docs/for-app-authors/appdata-guidelines/quality-guidelines/).

### LINKED ISSUE #632: Non-maximized appstream screenshots
In order to qualify for the new curation we want to introduce for Flathub this app needs non-maximized screenshots with shadows, rounded corners, transparency, etc. (as produced by the window screenshot feature in GNOME Shell).

tchx84:
Closes #612
Closes #622
Closes #632


tchx84:
@bertob is this better?

razzeee:
Technically these still seem to be above 1000px width

bertob:
The window itself is 1000x700, the rest is shadow.

These look good to me :+1:

razzeee:
We should probably clarify that in the guidelines then, I would have expected the image size to count here, as everything else is hard to figure out/track.

tchx84:
FWIW, the [guidelines](https://docs.flathub.org/docs/for-app-authors/appdata-guidelines/quality-guidelines/#reasonable-window-size) explicitly refers to the "window size".

> In order to ensure that text and interface elements are visible scaled-down in the app store UI **the window size** should be 1000x700 pixels or smaller (2000x1400 for HiDPI).

razzeee:
Your right, but our tooling kinda bites us then, as we write the size of the images to the view, when moderating

image

See screenshot sizes below the screenshot

tchx84:
> Your right, but our tooling kinda bites us then, as we write the size of the images to the view, when moderating
> See screenshot sizes below the screenshot

I understand what you're saying now.

These guidelines are referring to the contents of the image so, unless those "rules" can be "coded" into that moderation tool, it will certainly create confusion.

From an app developer perspective, it's annoying to be joggling with the window size to hopefully produce a screenshot that doesn't extend over some fixed size. Maybe is here where we need better tooling (to ease taking these screenshots, and leave some fixed size on the moderation tool).

This time I had to resort to the gtk-inspector to take these screenshots which is already eye'rolly.

I wonder what other people are doing...

razzeee:
> Maybe is here where we need better tooling (to ease taking these screenshots, and leave some fixed size on the moderation tool).

Have been thinking this a bunch of times the last months, but probably can't afford to get that going. Could be a nice gsoc project, if it finds a mentor.

bertob:
Long-term the only sustainable solution here is probably making shadows server-side :see_no_evil: 

Personally I've not found the sizes shown on the page helpful, both because of shadows and since they're often wrong (maybe because of HiDPI, maybe because they're being re-encoded or something, not sure). In practice I'm pretty good at telling if something is roughly on the level, and in the rare case where I'm not sure I go measure it in Inkscape. Not ideal, but sort of works for now :)

PR #658, opened 2024-02-18
String fixes

jamiethecat:


tchx84 (changes requested):
Hi @MonsterObserver, unfortunately I can't merge these improvements as they are now. Is just too much at once, as it would break all translations and trigger a lot of work (burden) for translators.

I suggest that you break down this work and simply start with the documentation for now, which won't break anything, and we see where to go from there.

PR #665, opened 2024-03-10
Update Hungarian translation

rdbende:
Update Hungarian strings to be correct in the context of Flatseal

### DIFF
--- a/po/hu.po
+++ b/po/hu.po
@@ -2,20 +2,22 @@
 # Copyright (C) 2020 THE flatseal'S COPYRIGHT HOLDER
 # This file is distributed under the same license as the flatseal package.
 # Automatically generated, 2020.
+# Dévényi Benedek <rdbende@proton.me>, 2024.
 #
 msgid ""
 msgstr ""
 "Project-Id-Version: flatseal\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2024-01-18 10:37-0300\n"
-"PO-Revision-Date: 2020-02-01 03:47-0300\n"
-"Last-Translator: Automatically generated\n"
-"Language-Team: none\n"
+"PO-Revision-Date: 2024-03-10 12:44+0100\n"
+"Last-Translator: Dévényi Benedek <rdbende@proton.me>\n"
+"Language-Team: Hungarian <>\n"
 "Language: hu\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+"Plural-Forms: nplurals=2; plural=(n != 1)\n"
+"X-Generator: Gtranslator 45.3\n"
 
 #. TRANSLATORS: Don't translate this text
 #: data/com.github.tchx84.Flatseal.desktop.in:4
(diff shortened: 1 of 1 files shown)


tchx84:
Hey @rdbende , thanks for taking the time to work on this... @ovari want to take a quick look?

ovari:
Thank you @rdbende for your translation updates and taking over the Hungarian translations.

If you are interested, below are a few pointers for your toolkit.

### Localization databases

1. Visit https://learn.microsoft.com/en-us/globalization/reference/microsoft-language-resources
2. Click on `Microsoft UI String Search` link in the [UI Strings](https://learn.microsoft.com/en-us/globalization/reference/microsoft-language-resources#ui-strings) section

Microsoft: https://learn.microsoft.com/en-us/globalization/reference/microsoft-language-resources
Mozilla: https://transvision.mozfr.org/
KDE: https://l10n.kde.org/dictionary/search-translations.php

### sandbox

Literal translation: `sandbox` → `homokozó` (incorrect?)
Microsoft software terminology translation: `sandbox` → `védőfal`
Mozilla software terminology translation: `sandbox` → `homokozó` (so perhaps the literal translation is not incorrect?!)
KDE: `sandbox` → `helyi adattár` (similar in context to `védőfal`)

### move …

Should `Move up` be `Mozgatás felfelé` (based on Microsoft and Mozilla?).

Should the other `Move …` should also be updated?

### About «programName»

Menu bar, _Help, _About «programName»
Menüsor, _Súgó, _A «programNév» névjegye

Reasons to keeping the keyboard shortcut for `_About Flatseal` in Hungarian as `_A Flatseal névjegye`:
1. The keyboard shortcut for English and Hungarian will be the same, i.e. both `_A`
2. The keyboard shortcut for Hungarian will be consistent with other programs which are in Hungarian, including but not limited to: `_A LibreOffice névjegye`, `_A Thunderbird névjegye`

### Main Menu

1. Is there only one menu? If so, perhaps the English string should be changed from `Main Menu` to `Menu`?
2. Does this relate to a tooltip? If so, perhaps the English string should include `Open ` to become `Open menu`?

### Controller Area Network

Cf. https://hu.wikipedia.org/wiki/CAN-busz

If you want this to be Hungarian, is a hyphen `-` required?
`CAN-busz`

If you really want to include the English in the translation, it might be better to write (please note the [En Dash (–)](https://unicodeplus.com/U+2013) between English – Hungarian):
`CAN (Controller Area Network – Vezérlő terület hálózati) -busz`

Perhaps this is a discussion that can happen between @rdbende and @tchx84

Thank you once again

rdbende:
> ### sandbox

According to [its Wikipedia page](https://hu.wikipedia.org/wiki/Homokoz%C3%B3_(biztons%C3%A1g)) it's called `homokozó`, and other Hungarian articles I've read about Flatpak call it `homokozó` as well, so I think in this case it's right to use the literal translation.


> ### move …

It should be `Mozgatás` if it would be an action to move an element, but in this context it moves the focus, so `ugrás` is a more appropriate translation.


> ### Main Menu

Yes, it is a tooltip for the application menu. Apps label it pretty inconsistently (I've seen both `Open Menu` and `Main Menu`), but the GNOME HIG seems to recommend using `Main Menu`, so I'll change the translation to `Főmenü`.


> ### About «programName»

I've yet to see an exclusively Linux app, that uses `a` as the mnemonic for the about action in the Hungarian translation, but it's true that most cross-platform apps use it. Most GNOME apps I tried on my system however use `n` as the mnemonic, and it's better to be consistent in that. 

I don't know about KDE, but the GNOME HIG has this about mnemonic keys:

> Additionally, if another letter provides a better association (for example: “x” in “Extra Large”), consider using that letter instead"[^1],

and I think this is a case for it, since the article `a` is too generic to be something that's easily memorable.


> ### Controller Area Network

I did a bit of research on this, and I found no references of `Vezérlő terület halózati busz` in Hungarian, it's always referred to as `CAN-busz`.
Therefore if someone needs this option, they're going to look for it in its Enligsh name. However, since the CAN abbreviation is more common in Hungarian than its full name, the best translation would be `CAN (Controller Area Network) -busz`. I'll update that.


[^1]: [GNOME HIG](https://developer.gnome.org/hig/guidelines/keyboard.html#access-keys)

rdbende:
And btw, thanks for the resources!

tchx84 (approved):

PR #694, opened 2024-05-08
aboutDialog: Add @Scrambled777 to credits

tchx84:


tchx84:
@Scrambled777 let me know if you prefer another alias or name.

Scrambled777:
> @Scrambled777 let me know if you prefer another alias or name.

Hi @tchx84 , `Scrambled777` is perfect and thanks. Cheers.

PR #705, opened 2024-07-13
variables: Relax variable name validation

### LINKED ISSUE #700: Pavucontol environment variable issue
Hi, recently I have a problem with setting up pavucontrol environment variables with Flatseal

Here's the issue from pavucontrol GitHub - https://github.com/flathub/org.pulseaudio.pavucontrol/issues/21

Environment variable `PULSE_PROP_media.category=Manager` (which is needed for Pavucontrol) is not considered valid by Flatseal, but according to one of pavucontrol package contributors (in last message from pavucontrol issue above) this variable should be valid and this is an issue on Flatseal side

Right now it's not convenient to use Flatseal with pavucontrol. Can we fix this?

tchx84:
Fixes #700

### DIFF
--- a/src/models/variables.js
+++ b/src/models/variables.js
@@ -23,7 +23,7 @@ const {GObject} = imports.gi;
 const {FlatpakSharedModel} = imports.models.shared;
 const {FlatsealOverrideStatus} = imports.models.overrideStatus;
 
-var VAR_REGEXP = /^\w+=[\S ]+$/;
+var VAR_REGEXP = /^[^;\s]+=[\S ]+$/;
 
 var FlatpakVariablesModel = GObject.registerClass({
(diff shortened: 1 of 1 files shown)


tests/src/testModels.js
```diff
@@ -803,7 +803,7 @@ describe('Model', function() {
 
         expect(permissionsDefault.variables).toEqual('TEST=yes');
 
-        permissionsDefault.set_property('variables', 'TEST=yes;TE ST=no');
+        permissionsDefault.set_property('variables', 'TEST=yes;TE;ST=no');
```
tchx84: note to future self; find a better way to serialize/deserialize this list, in a way that doesn't clash with the actual values...

PR #728, opened 2024-10-07
Add screenshot portal permission row

PolyMeilex:
~Some apps (like flameshot https://github.com/flameshot-org/flameshot/issues/2868#issuecomment-2397392428) require users to set the screenshot portal permission manually. Flatseal seems like a right place for this dial.~

This allows users to revoke previously given permission. (it was previously posible in gnome control center, but having it in one place in Flatseal would be cool too)

image

### DIFF
--- a/src/models/portals.js
+++ b/src/models/portals.js
@@ -79,6 +79,7 @@ var FlatpakPortalsModel = GObject.registerClass({
         this._devicesspeakersSupported = null;
         this._devicescameraSupported = null;
         this._locationlocationSupported = null;
+        this._screenshotscreenshotSupported = null;
 
         this._backgroundbackgroundReason = '';
         this._notificationsnotificationReason = '';
@@ -86,6 +87,7 @@ var FlatpakPortalsModel = GObject.registerClass({
         this._devicesspeakersReason = '';
         this._devicescameraReason = '';
         this._locationlocationReason = '';
+        this._screenshotscreenshotReason = '';
 
         this._info = info.getDefault();
         this._appId = '';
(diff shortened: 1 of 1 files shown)


tchx84:
hey @PolyMeilex , I scrolled through that thread and let me see if I understand it... Flameshot can't rely on the regular portal UI because it doesn't have a window?

> Some apps (like flameshot https://github.com/flameshot-org/flameshot/issues/2868#issuecomment-2397392428) require users to set the screenshot portal permission manually. Flatseal seems like a right place for this dial.

Unfortunately, I don't think this will work. Preventing this use-case was a very explicit decision.

By "this use-case" I mean having Flatseal creating tables that haven't been initialized by portals themselves by properly interacting with the application. Some of the reasons are:
* To avoid confusion for how portals work, e.g., an application won't be granted camera access just because somebody set a permissions store value, since it has to be properly implemented in the app first.
* To limit Flatseal's ability to screw up the permission store.

So, Flatseal can't be used to set a portal that hasn't been used yet.

PolyMeilex:
> Flameshot can't rely on the regular portal UI because it doesn't have a window?

As far as I understand, GNOME portal requires the app window to have focus for the permission dialog to spawn. (and Flameshot appears to not have a typical toplevel window at all)

I don't care about this that much tho, just thought that it might be one of the uses, if it's not, that's fine as well. 

I'm still interested in using Flatseal to revoke previously given screenshot permissions tho, any ideas why Flatseal allows me to set this permission for apps that never called the portal? (Same thing happens with Notification/Background permissions)

Arcitec:
This is something that either GNOME or Flameshot has to fix.

The problem is that Flameshot attempts to use the screenshot portal while Flameshot doesn't have any application window. This suppresses the permission popup dialog on GNOME.

So Flameshot should implement code to generate a GUI window and ask for permission once. They could add a button to their config GUI window when they detect being run inside a Flatpak, for example. Or they may even be able to query Flatpak to see if they have the permission, and if not, generate a GUI window which then asks for the permission.

Either way, I agree that Flatseal should not be writing permissions that apps didn't register for!

Edit: But I agree with the previous comment that it would be good if Flatseal can detect and offer a way to REVOKE the Screenshot permission.

PR #730, opened 2024-10-12
implementation of translation of bengali

olumolu:


po/bn.po
```diff
@@ -0,0 +1,608 @@
...
+msgid "seal;sandbox;override;"
+msgstr "seal;sandbox;override;"
+
+#: data/com.github.tchx84.Flatseal.appdata.xml.in:12
+msgid """Flatseal is a graphical utility to review and modify permissions from your ""Flatpak applications."
+msgstr "ফ্ল্যাটসিল একটি গ্রাফিক্যাল ইউটিলিটি যা আপনার ফ্ল্যাটপ্যাক অ্যাপ্লিকেশনগুলির অনুমতি পর্যালোচনা এবং পরিবর্তন করতে ব্যবহৃত হয়।"
+
+#: data/com.github.tchx84.Flatseal.appdata.xml.in:21
+msgid "The Flatseal main window"
+msgstr "ফ্ল্যাটসিল প্রধান উইন্ডো"
```
tchx84: I am not familiar with Bengali, so I am not sure it's customary to translate the application name like this.

po/bn.po
```diff
@@ -0,0 +1,608 @@
...
+msgid "Portal Version"
+msgstr "পোর্টাল সংস্করণ"
+
+#: src/widgets/globalInfoViewer.ui:81
+msgid "Changes Path"
+msgstr "পথ পরিবর্তন"
+
+#: src/widgets/menu.ui:7
+msgid "_Help"
+msgstr "_সাহায্য"
```
tchx84: I wonder if the mnemonic will work here, e.g., I noticed gnome-shell Bengali translation simply ignores the `_` prefix. Maybe we should also do the same here.

tchx84 (changes requested):
Hi @olumolu , I left two minor comments.

 Thanks for working on this and apologies for the delay.

po/bn.po
olumolu: Sure i can left that in english

po/bn.po
olumolu: I can leave that in english if You want I don't know that gnome shell ignore that `_`

tchx84: Not sure either, but I would not suggest to leave it in English. Perhaps just remove the `_` prefix it, and change all other similar strings to keep it consistent.

po/bn.po
tchx84: I would suggest you look at gnome-shell (and maybe other apps) [translation](https://gitlab.gnome.org/GNOME/gnome-shell/-/blob/main/po/bn.po?ref_type=heads) and follow whatever it's done there.

olumolu: I have done that for alpaca and all we have translated all lines

po/bn.po
```diff
@@ -0,0 +1,608 @@
...
+msgstr "রানটাইম"
+
+#: src/widgets/busNameRow.js:35 src/widgets/pathRow.js:73
+#: src/widgets/relativePathRow.js:33 src/widgets/variableRow.js:33
+msgid "This is not a valid option"
+msgstr "এটি একটি বৈধ বিকল্প নয়"
+
+#: src/widgets/detailsButton.js:60
+msgid "_Show Details"
+msgstr "_বিস্তারিত দেখুন"
```
tchx84: Same here.

po/bn.po
```diff
@@ -0,0 +1,608 @@
...
+msgid "Changes Path"
+msgstr "পথ পরিবর্তন"
+
+#: src/widgets/menu.ui:7
+msgid "_Help"
+msgstr "সাহায্য"
+
+#: src/widgets/menu.ui:11
+msgid "_Documentation"
+msgstr "_ডকুমেন্টেশন"
```
tchx84: Same here.

po/bn.po
```diff
@@ -0,0 +1,608 @@
...
+
+#: src/widgets/menu.ui:15
+msgid "_Keyboard Shortcuts"
+msgstr "_কীবোর্ড শর্টকাট"
```
tchx84: Same here.

po/bn.po
```diff
@@ -0,0 +1,608 @@
...
+
+#: src/widgets/menu.ui:19
+msgid "_About Flatseal"
+msgstr "_ফ্ল্যাটসিল সম্পর্কে"
```
tchx84: Same here.

po/bn.po
```diff
@@ -0,0 +1,608 @@
...
+msgstr "সেট করা হয়নি"
+
+#: src/widgets/relativePathRow.js:84
+msgid "Default paths can't be removed"
+msgstr "ডিফল্ট পথগুলি সরানো যাবে না"
+
+#: src/widgets/resetButton.js:37 src/widgets/resetButton.js:55
+#: src/widgets/resetButton.js:62 src/widgets/window.js:106
+msgid "_Reset"
+msgstr "_রিসেট করুন"
```
tchx84: Same here.

po/bn.po
```diff
@@ -0,0 +1,608 @@
...
+msgid "Find previous"
+msgstr "আগেরটি অনুসন্ধান করুন"
+
+#: src/widgets/window.js:98
+msgid "Permissions have been reset"
+msgstr "অনুমতিগুলি রিসেট করা হয়েছে"
+
+#: src/widgets/window.js:99
+msgid "_Undo"
+msgstr "_আনডু"
```
tchx84: Same here.

po/bn.po
```diff
@@ -0,0 +1,608 @@
...
+msgid """Flatseal is a graphical utility to review and modify permissions from your ""Flatpak applications."
+msgstr "ফ্ল্যাটসিল একটি গ্রাফিক্যাল ইউটিলিটি যা আপনার ফ্ল্যাটপ্যাক অ্যাপ্লিকেশনগুলির অনুমতি পর্যালোচনা এবং পরিবর্তন করতে ব্যবহৃত হয়।"
+
+#: data/com.github.tchx84.Flatseal.appdata.xml.in:21
+msgid "The Flatseal main window"
+msgstr "Flatseal প্রধান উইন্ডো"
+
+#: data/com.github.tchx84.Flatseal.appdata.xml.in:25
+msgid "Flatseal showing filesystem permissions"
+msgstr "ফ্ল্যাটসিল ফাইল সিস্টেম অনুমতি দেখাচ্ছে"
```
tchx84: Whether you keep it in English or Bengali, just be consistent through all the document.

po/bn.po
```diff
@@ -0,0 +1,608 @@
...
+
+#: data/com.github.tchx84.Flatseal.appdata.xml.in:29
+msgid "Flatseal showing global overrides"
+msgstr "ফ্ল্যাটসিল গ্লোবাল ওভাররাইড দেখাচ্ছে"
```
tchx84: Be consistent here.

po/bn.po
tchx84: Also be consistent here.

tchx84 (changes requested):
I still see places where you translate Flatseal and places where you don't translate Flatseal. I don't really know what's the correct way of doing it for Bengali,  but I think it should at least be consistent. So please make it consistent, no matter what you choose.

PR #744, opened 2024-12-23
Move appstream metadata to modern location.

danialbehzadi:
Appstream metadata was on legacy location (/usr/share/appdata). It should be moved to the modern location (/usr/share/metainfo).

https://wiki.debian.org/AppStream/Guidelines


data/meson.build
```diff
@@ -19,7 +19,7 @@ appstream_file = i18n.merge_file(
   output: 'com.github.tchx84.Flatseal.appdata.xml',
   po_dir: '../po',
   install: true,
-  install_dir: join_paths(get_option('datadir'), 'appdata')
+  install_dir: join_paths(get_option('datadir'), 'metainfo')
```
tchx84, line 22: Perhaps we should also rename the file to metainfo.xml.

tchx84 (changes requested):

tchx84:
@danialbehzadi fixed it myself, thanks.

PR #745, opened 2024-12-25
Typo in french translation

dynamiclinking:
The typo made that it read sand bag instead of sand box.

I don't know if this is the correct way to contribute to the translation of flatseal.


tchx84 (approved):

tchx84:
@dynamiclinking nice catch, thank you!

PR #751, opened 2025-01-22
Implement a --show-app CLI flag

heliguy4599:
## What this adds:
`flatseal --show-app org.app.Name`
- Adds a new --show-app command-line flag, which takes a Flatpak's appID.
- If the supplied ID matches an installed app, the UI will switch to showing that app.
- If no match is found, nothing happens. 
- The flag can be invoked at startup of Flatseal, or while Flatseal is running, and will behave the same.

There is no way to show the Global section, however I can add this if it is desired. Also, I can change the behavior of a non-matching provided ID, if you'd like that to behave differently than doing nothing.

The new flag shows up in the default --help output, with a description.

I believe I have followed Flatseal's coding style to the best of my abilities, but I can always refine what I've added!

## Why I think this should be added:
I believe it would be handy for Flatseal to be able to be opened to a specific application, to allow it to be called from other apps or scripts. I am the developer of [Warehouse](https://github.com/flattool/warehouse), and I would like to be able to launch Flatseal to a specific app via a button in the UI.

### DIFF
--- a/src/application.js
+++ b/src/application.js
@@ -43,7 +43,7 @@ var FlatsealApplication = GObject.registerClass({
             null,
             GLib.OptionFlags.NONE,
             GLib.OptionArg.STRING,
-            'A custom flag for my app',
+            'Show a specific application entry, by providing its ID',
             null);
 
         this.connect("command-line", this._cliArgHandler.bind(this));
--- a/src/application.js
+++ b/src/application.js
@@ -19,7 +19,7 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-const {GObject, Gtk, Gio, Adw} = imports.gi;
+const {GObject, Gtk, Gio, GLib, Adw} = imports.gi;
 
 const {FlatsealWindow} = imports.widgets.window;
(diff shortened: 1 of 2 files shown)


tchx84:
hey @heliguy4599 , I will take a look at this ASAP. Thanks.

heliguy4599:
Great to hear! You're welcome, and thank you for this wonderful app itself!

tchx84 (changes requested):
Apologies for the big delay.

This looks OK, but wouldn't it better to expose this functionality as a service?

So that, instead of having to call flatseal via CLI (which is an unlikely use case for actual people), one could simply activate Flatseal with the proper appId via DBus (which is what programs should be doing, like your use case).

I am pretty sure there is a GApplication/GAction magic that does all the heavy lifting here.

tchx84:
In case there's still interest, the idea would be for other apps to invoke this service similarly to how Flatseal [invokes](https://github.com/tchx84/Flatseal/blob/master/src/widgets/detailsButton.js#L81) GNOME software today.

tchx84:
Superseded by https://github.com/tchx84/Flatseal/pull/799

PR #769, opened 2025-04-29
Update fr.po

badbrocoli:
"bac à sable" is the correct translation for sandbox.


tchx84:
Oh, it seems there were more entries than the one fixed in https://github.com/tchx84/Flatseal/pull/745

tchx84:
@badbrocoli thanks!

PR #773, opened 2025-05-16
feat: add device=usb permission toggle

HastD:
Raw USB device access can be granted using `device=usb` as of flatpak version 1.15.11:
https://docs.flatpak.org/en/latest/sandbox-permissions.html#device-access

### DIFF
--- a/tests/src/testModels.js
+++ b/tests/src/testModels.js
@@ -33,7 +33,7 @@ const {
 
 setup();
 
-const _totalPermissions = 40;
+const _totalPermissions = 41;
 
 const _basicAppId = 'com.test.Basic';
 const _basicNegatedAppId = 'com.test.BasicNegated';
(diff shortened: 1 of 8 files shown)


tests/src/testModels.js
```diff
@@ -302,6 +305,7 @@ describe('Model', function() {
         expect(permissionsDefault.devices_input).toBe(true);
         expect(permissionsDefault.devices_kvm).toBe(true);
         expect(permissionsDefault.devices_shm).toBe(true);
+        expect(permissionsDefault.devices_usb).toBe(true);
         expect(permissionsDefault.devices_all).toBe(true);
         expect(permissionsDefault.features_bluetooth).toBe(true);
         expect(permissionsDefault.features_devel).toBe(true);
```
tchx84, line 311: This is failing because there are a few tests that "count" the number the "active" permissions (depending on different Flatpak versions), so you need to increase the total number of permissions (in this same file) and also increase that number in the tests where the version of Flatpak is "very new".

tchx84 (changes requested):
hey @HastD , thanks for working on this. Just a comment regarding tests that are failing.

tests/src/testModels.js
HastD: I think it should be fixed now.

tchx84 (approved):

tchx84:
thanks @HastD !

PR #788, opened 2025-07-09
fix: Replace deprecated symbolic icon

rivenirvana:
Since GNOME 48, the `emblem-*-symbolic` icons have been [deprecated](https://gitlab.gnome.org/GNOME/adwaita-icon-theme/-/merge_requests/78/) and it is not backported to adwaita-icon-theme-legacy. This replaces it with a similar-looking checkmark icon so those that do not use an extensive custom icon theme (just base Adwaita) can see it.

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -36,7 +36,7 @@ row .content .path .info {
 row .content .relative.path .info,
 row .content .bus .info,
 row .content .variable .info {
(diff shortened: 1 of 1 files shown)


tchx84 (approved):

tchx84:
@rivenirvana thanks!

PR #800, opened 2025-08-30
dbus: Support activating Flatseal with a given appId

### LINKED ISSUE #782: Allow specifying an initial appid to jump to in a command line invokation of Flatseal
Hello, I'm currently working on this project https://github.com/kolunmi/bazaar

I have a button labeled "Manage Permissions" next to each app ref on the installed page which just opens flatseal, but unfortunately I have not found a way to get flatseal to show the specific application selected. Is it possible you could add this functionality? Thank you for your time.

tchx84:
This feature has been requested a few times now by the developers of projects like Warehouse, Bazaar, and others.

The following is an example of how to use this API:

```
gdbus call \
  --session \
  --dest com.github.tchx84.Flatseal \
  --object-path /com/github/tchx84/Flatseal \
  --method org.gtk.Actions.Activate \
  show "[<string 'org.gnome.Builder'>]" "{}"
```

Closes #782

### DIFF
--- a/data/com.github.tchx84.Flatseal.desktop.in
+++ b/data/com.github.tchx84.Flatseal.desktop.in
@@ -12,3 +12,4 @@ Icon=com.github.tchx84.Flatseal
 # Translators: Do NOT translate or transliterate this text (these are enum types)!
 X-Purism-FormFactor=Workstation;Mobile;
 Keywords=seal;sandbox;override;
+DBusActivatable=true
--- a/data/com.github.tchx84.Flatseal.service.in
+++ b/data/com.github.tchx84.Flatseal.service.in
@@ -1,0 +1,3 @@
+[D-BUS Service]
+Name=com.github.tchx84.Flatseal
+Exec=@prefix@/@bindir@/com.github.tchx84.Flatseal --gapplication-service
--- a/data/meson.build
+++ b/data/meson.build
@@ -41,4 +41,15 @@ if compile_schemas.found()
   )
 endif
 
+conf = configuration_data()
+conf.set('prefix', get_option('prefix'))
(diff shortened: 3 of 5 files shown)


tchx84:
@heliguy4599 @kolunmi I think I will go with this implementation.. 

As I mentioned before, this can be used similarly to [this](https://github.com/tchx84/Flatseal/blob/master/src/widgets/detailsButton.js#L81).

For context, a few advantages over the CLI command:

1. You can inspect the session dbus to see if Flatseal is available (and disable related UI when not).
2. You can inspect the session dbus to see if the installed Flatseal version supports this new "show" action (and disable related UI when not).
3. And only then allow your users to call this action over dbus.

All the above allows a better UX than just "spawn flatseal and if something fails, well, too bad".

Additionally, makes the relationship more transparent as your apps will need a specific static permission to talk to Flatseal over dbus... reducing the amount of scary "flatpak spawn"s.

kolunmi:
Hi @tchx84  I sincerely apologize for not testing this yesterday, I had a lot going on.

I built Flatseal locally and tested this new feature. It works perfectly! I appreciate you putting thought into the quality of this solution (It will certainly be useful in Bazaar). Thus, if you were looking for approval from me, I'd say go ahead and merge this pr. Thanks!

heliguy4599:
Hello @tchx84! Thank you so much for this thoughtful approach! Building and testing Flatseal locally, the feature works great here too! It'll be very useful for Warehouse. I appreciate it very much, thanks a bunch for adding a feature like this! Like with Kolunmi, I also say go ahead merge! Take care and have a great day!

tchx84:
Thank you both for testing it :)

PR #832, opened 2025-11-20
Add Vietnamese translation

hthienloc:


po/vi.po
```diff
@@ -0,0 +1,614 @@
...
+msgid "Changes Path"
+msgstr "Đường Dẫn Thay Đổi"
+
+#: src/widgets/menu.ui:7
+msgid "_Help"
+msgstr "_Trợ giúp"
+
+#: src/widgets/menu.ui:11
+msgid "_Documentation"
+msgstr "_Tài Liệu"
```
tchx84: These mnemonics are supposed to be unique, so try to use another character (e.g., "Tài _Liệu").

tchx84 (changes requested):

po/vi.po
hthienloc: Can you check again? I've just updated the translation.

PR #839, opened 2026-01-07
Add Kernewek translation

linfindel:


po/kw.po
```diff
@@ -0,0 +1,618 @@
...
+msgid "_Help"
+msgstr "_Gweres"
+
+#: src/widgets/menu.ui:11
+msgid "_Documentation"
+msgstr "_Kowethlyver"
+
+#: src/widgets/menu.ui:15
+msgid "_Keyboard Shortcuts"
+msgstr "_Kesunyansow Bysowek"
```
tchx84: Mnemonics are supposed to be "unique" within the same context (e.g., the dialog where this text is shown), so you can't have "_Kowethlyver" and "_Kesunyansow", using the same letter. You can change it do another letter like, "Kesunyansow _Bysowek", so the mnemonics will be "B" and different.

tchx84 (changes requested):
@linfindel looks good, just a small detail.

po/kw.po
linfindel: Should be fixed now

PR #840, opened 2026-01-09
Use new GTK session save/restore API

DaPigGuy:
GTK is now able to handle saving/restoring sessions. It automatically handles restoring window size/fullscreen/maximized state. On Wayland, it integrates with the (currently experimental) session management protocol to restore the window position, which is probably the biggest motivator for this, although this is currently not ready when [running under Flatpak](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/8980)? GTK does auto saves every 15s and on app closing.

See for more info: [GNOME initiative](https://gitlab.gnome.org/GNOME/Initiatives/-/issues/62)

API is not yet stable, so let's wait til GNOME 50


tchx84:
@DaPigGuy thanks for working on this, I will take a look this week and let's merge it when it's stable

src/application.js
```diff
@@ -102,6 +103,26 @@ var FlatsealApplication = GObject.registerClass({
...
+    vfunc_restore_window(reason, state) {
+        this._window = new FlatsealWindow(this);
+
+        /* Avoid restoring selected app if restoring from a crash in order to avoid potential crash loop */
+        if (reason !== Gtk.RestoreReason.RECOVER) {
+            const selectedAppId = state.lookup_value('selected-app-id', new GLib.VariantType('s'))?.get_string()[0];
+            this._window.showApplication(selectedAppId);
+        }
+
+        /* Only restore the search entry for crash recover or session restore */
```
tchx84: Let's leave this search-entry restore behavior out for now, and focus on the existing behaviors. Making this PR simple as possible.

tchx84 (changes requested):

src/application.js
DaPigGuy: Sure

DaPigGuy:
Looks like this won't be happening for GNOME 50 :sweat_smile: 

https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/9395

tchx84:
@DaPigGuy any news on this now that GNOME 50 is out? :)

PR #844, opened 2026-02-09
Added Portuguese Portugal translation

rottenpants466:


po/pt.po
```diff
@@ -0,0 +1,618 @@
...
+msgid "_Documentation"
+msgstr "_Documentação"
+
+#: src/widgets/menu.ui:15
+msgid "_Keyboard Shortcuts"
+msgstr "_Atalhos do teclado"
+
+#: src/widgets/menu.ui:19
+msgid "_About Flatseal"
+msgstr "_Acerca do Flatseal"
```
tchx84, line 386: Mnemonics are supposed to be "unique" within the same context (e.g., the dialog where this text is shown), so you can't have "_Atalhos" and "_Acerca", using the same letter. You can change it do another letter like, "Acerca do _Flatseal", so the mnemonics will be "F" and different.

tchx84 (changes requested):
hey @rottenpants466 , thanks for working on this. This looks good, there's only a small detail to fix.

PR #850, opened 2026-03-08
fix: Handle malformed custom installation config files gracefully

YusufErdemK:
_parseCustomInstallation() did not wrap keyFile.load_from_file() in a
try-catch block. If /etc/flatpak/installations.d/ contained a malformed,
binary, or otherwise unreadable .conf file, Flatseal would crash entirely
instead of skipping the bad file.

Also simplify the version check in portals.js for readability:
  Before: this._proxy.version >= SUPPORTED_SERVICE_VERSION === false
  After:  this._proxy.version < SUPPORTED_SERVICE_VERSION
(functionally equivalent, but clearer)

### DIFF
--- a/src/models/portals.js
+++ b/src/models/portals.js
@@ -260,7 +260,7 @@ var FlatpakPortalsModel = GObject.registerClass({
 
         this._setup();
 
-        if (this._proxy === null || this._proxy.version >= SUPPORTED_SERVICE_VERSION === false) {
+        if (this._proxy === null || this._proxy.version < SUPPORTED_SERVICE_VERSION) {
             this[`_${table}${id}Reason`] = _('Requires permission store version 2 or newer');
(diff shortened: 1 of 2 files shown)


src/models/portals.js
```diff
@@ -260,7 +260,7 @@ var FlatpakPortalsModel = GObject.registerClass({
 
         this._setup();
 
-        if (this._proxy === null || this._proxy.version >= SUPPORTED_SERVICE_VERSION === false) {
+        if (this._proxy === null || this._proxy.version < SUPPORTED_SERVICE_VERSION) {
```
tchx84, line 263: This seems right, but I would prefer not to include random unrelated changes. Let's merge the actual fix first?

tchx84 (changes requested):
they @YusufErdemK , thanks for the fix. Please see my comment.

PR #851, opened 2026-03-08
Feat/modified permissions indicator

### LINKED ISSUE #837: [Feature Request] Visual indicator for modified permissions in applications list
### Problem

Currently, Flatseal only shows modified permissions when viewing an app's details. There's no way to identify which apps have user-modified permissions from the main app list view.

### Solution

Add a visual indicator in the app list to show which apps have permissions that differ from their defaults. Options:

1. Icon badge (e.g., small dot or asterisk next to app name)
2. Different text color or styling
3. Additional column showing "Modified" status
4. Filter/sort option to show only modified apps

### Use Case

Users managing multiple Flatpak applications need to quickly identify which apps they've customized without opening each one individually.

KafilatAdeleke:
Closes #837
Adds a visual indicator in the application list for apps with user-modified permissions. The badge appears on both individual app rows and the global row. Detection combines portal-level changes and file-based overrides across all override paths.

### DIFF
--- a/src/models/applications.js
+++ b/src/models/applications.js
@@ -389,12 +389,12 @@ var FlatpakApplicationsModel = GObject.registerClass({
         return GLib.build_filenamev([this._getBundlePathForAppId(appId), 'metadata']);
     }
 
-    getOverridesPaths() {
-        return this._getInstallationsPaths().map(path => GLib.build_filenamev([path, 'overrides']));
-    }
-
     get userPath() {
         return this._getUserPath();
+    }
+
+    get userOverridesPath() {
+        return GLib.build_filenamev([this._getUserPath(), 'overrides']);
     }
 
     shutdown() {
--- a/src/models/applications.js
+++ b/src/models/applications.js
@@ -389,6 +389,10 @@ var FlatpakApplicationsModel = GObject.registerClass({
         return GLib.build_filenamev([this._getBundlePathForAppId(appId), 'metadata']);
     }
 
+    getOverridesPaths() {
+        return this._getInstallationsPaths().map(path => GLib.build_filenamev([path, 'overrides']));
+    }
+
     get userPath() {
         return this._getUserPath();
     }
--- a/src/models/portals.js
+++ b/src/models/portals.js
@@ -384,20 +384,6 @@ var FlatpakPortalsModel = GObject.registerClass({
         return false;
     }
 
-    getAppsWithPortalChanges() {
-        const apps = new Set();
-        for (const [, permission] of Object.entries(this.getPermissions())) {
-            if (!this.isSupported(permission.table, permission.id))
-                continue;
-
-            const [appIds] = this.safeLookUp(permission.table, permission.id);
-            if (appIds !== null)
-                Object.keys(appIds).forEach(appId => apps.add(appId));
-        }
-
-        return apps;
-    }
-
     saveToKeyFile() {
 
         /* this backend has no file */
--- a/src/models/portals.js
+++ b/src/models/portals.js
@@ -384,7 +384,7 @@ var FlatpakPortalsModel = GObject.registerClass({
         return false;
     }
 
-    getAppsWithChanges() {
+    getAppsWithPortalChanges() {
         const apps = new Set();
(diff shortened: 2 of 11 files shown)


YusufErdemK (commented):
i didnt found any problem, thats a good pr

src/models/portals.js
```diff
@@ -384,6 +384,20 @@ var FlatpakPortalsModel = GObject.registerClass({
         return false;
     }
 
+    getAppsWithPortalChanges() {
```
tchx84: I been thinking about this change, and I don't think we should "track" portal changes. Here's why:

* Flatpak has two permissions models, static set by developers at build time (that aren't really meant to be modified) and dynamic set by users at run time (that are always supposed to be modified).
* Since dynamic permissions are always modified, it means this indicator will always be "on" for all the applications, and thus, it won't really serve its original purpose.

Therefore, I think we should limit this feature to only static permissions which aren't supposed to be modified all the time, and thus, have this indicator serve a purpose.

src/widgets/window.js
```diff
@@ -121,6 +121,11 @@ var FlatsealWindow = GObject.registerClass({
         this._applicationsListBox.set_sort_func(this._sort.bind(this));
         this._applicationsListBox.connect('row-activated', this._activateApplication.bind(this));
 
+        this._permissions.connect('changed', (model, overriden) => {
+            if (this._activatedRow)
+                this._activatedRow.changed = overriden;
+        });
```
tchx84: This would be gone.

src/widgets/window.js
```diff
@@ -155,12 +160,30 @@ var FlatsealWindow = GObject.registerClass({
...
+                const directory = Gio.File.new_for_path(overridesPath);
+                const enumerator = directory.enumerate_children('standard::name', Gio.FileQueryInfoFlags.NONE, null);
+                let fileInfo = enumerator.next_file(null);
+
+                while (fileInfo !== null) {
+                    changedApps.add(fileInfo.get_name());
+                    fileInfo = enumerator.next_file(null);
+                }
+            }
+        });
```
tchx84: Instead of this, wouldn't cleaner to have this logic in a separate model that uses Gio monitors to track which overrides exist and are modified? This way, it would cover the case where an external program removes an override.

tchx84 (changes requested):
hey @KafilatAdeleke , thanks for working on this. Please see my comments.

src/models/applications.js
```diff
@@ -389,6 +389,10 @@ var FlatpakApplicationsModel = GObject.registerClass({
         return GLib.build_filenamev([this._getBundlePathForAppId(appId), 'metadata']);
     }
 
+    getOverridesPaths() {
+        return this._getInstallationsPaths().map(path => GLib.build_filenamev([path, 'overrides']));
```
tchx84: Flatseal doesn't not support other overrides path besides userPath, so better not to monitor the other ones.

tchx84 (changes requested):
@KafilatAdeleke thanks, this is what I meant. Left one comment.

KafilatAdeleke:
Updated, now using a single userOverridesPath getter and simplified the model accordingly.

src/widgets/window.js
```diff
@@ -84,6 +84,15 @@ var FlatsealWindow = GObject.registerClass({
         this._permissions = permissions.getDefault();
         this._applications = applications.getDefault();
 
+        this._overrides = new overrides.FlatpakOverridesModel(
+            this._applications.userOverridesPath);
+        this._overrides.connect('changed', () => {
+            this._applicationsListBox.foreach(row => {
```
tchx84: This won't work unfortunately (plus, `foreach` doesn't exist). Check how to properly iterate these rows, there example in the same file.

More importantly, please do test your PR before requesting for another review.

tchx84 (changes requested):

KafilatAdeleke:
flatseal


Updated based on review feedback. Screenshot showing the indicator working across multiple apps. The badge appears only on apps with user-modified permissions and is absent on unmodified ones.

PR #852, opened 2026-03-11
appInfoViewer: Add Identifier field

### LINKED ISSUE #385: Critique: Application id under Application name in Application list is unclear
# Issue

My neurodivergent girlfriend commented "I do not know what these links are for".
She thought, it was a *url*.

image

## Solution
There are two solutions
1. Append a prefix of ``ID:`` to signify that the following is an application ID
2. Move application ID out from the application list, instead displaying it on the right side under the Application name, in the same list as "Version, Last Updated, Runtime..."

### LINKED ISSUE #781: Show full Application ID in the details pane
### Problem

When selecting a Flatpak with a long Application ID, the sidebar entry is truncated (see attached screenshot).
The Application ID is not shown anywhere else in the UI, so there is no way to view or copy the complete ID from Flatseal itself.

### Expected behavior

The main details pane—where Version, Last Updated, and Runtime are listed should also display the full Application ID of the selected Flatpak. This would:

- Allow users to easily read and copy long IDs.
- Keep the sidebar concise while still providing full information.
- Prevent confusion when multiple remotes provide similarly-named apps.

### Steps to reproduce

1. Install any Flatpak whose `application-id` is longer than the sidebar width (e.g. `org.cataclysmdda.CataclysmDDA`).
2. Open Flatseal.
3. Select the app in the sidebar.
   - The sidebar entry is ellipsised, and the main pane does not show the ID.

### Proposed solution

Add a new Application ID row to the header section of the details pane (below Version / Runtime). This row can include a copy-to-clipboard button, mirroring the UX used for Runtime in GNOME Software.

# Attached screenshot

Image

mmistika:
Displays full app ID in the app info view.
Closes #781
Closes #385
image_2026-03-11_21-25-28

### DIFF
--- a/src/widgets/appInfoViewer.js
+++ b/src/widgets/appInfoViewer.js
@@ -33,7 +33,7 @@ const styles = {
 var FlatsealAppInfoViewer = GObject.registerClass({
     GTypeName: 'FlatsealAppInfoViewer',
     Template: 'resource:///com/github/tchx84/Flatseal/widgets/appInfoViewer.ui',
-    InternalChildren: ['icon', 'name', 'author', 'version', 'released', 'runtime'],
+    InternalChildren: ['icon', 'name', 'author', 'version', 'identifier', 'released', 'runtime'],
     Properties: {
         compact: GObject.ParamSpec.boolean(
             'compact',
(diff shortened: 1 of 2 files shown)


tchx84 (changes requested):
Thanks for working on this, it looks good. Can you:

* Rebase this PR on top of the master branch, and push the updated branch?
* (since you're at it) Add the "Closes" keyboard for two issues?

Thanks.

tchx84:
@mmistika thanks!

PR #853, opened 2026-04-07
Drop ineffective cleanup commands

yakushabb:
They are unnecessary because no files are affected by these commands.


tchx84 (approved):

tchx84:
@yakushabb thanks

PR #854, opened 2026-04-17
Update GNOME runtime version from 49 to 50

fayasnoushad:
Update GNOME runtime version from 49 to 50

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -7,7 +7,7 @@ jobs:
   testing:
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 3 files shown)


tchx84 (changes requested):
@fayasnoushad thanks for working on this.

Before we can merge this, we need the following change in flatter merged so we can run tests in CI with the GNOME 50 runtime.

https://github.com/andyholmes/flatter/pull/203

tchx84:
Thanks for working on this but, since I am preparing an alternative to run these tests, I will be working on this here https://github.com/tchx84/Flatseal/pull/861

PR #861, opened 2026-05-14
Bump to GNOME 50

tchx84:
Plus, switch a custom and minimal GitHub action just to run the tests.


tchx84:
Note for myself (and for whoever else is checking this PR): I still need to clean up that custom action a bit more then I will merge it.

PR #879, opened 2026-06-10
models: Ignore conditional permissions when parsing overrides

Malika7188:
Flatpak 1.17+ introduced conditional permissions (e.g. `--socket-if=x11:!has-wayland`). Flatseal was not aware of this syntax, causing the conditional part to be dropped  and the override file to be written incorrectly.
  
This change makes Flatseal skip any permission value starting with `if:` when parsing override files, so the file remains valid. Supporting conditional permissions is out of scope for now.

Includes a unit test to verify the behavior.

### DIFF
--- a/src/models/permissions.js
+++ b/src/models/permissions.js
@@ -190,6 +190,8 @@ var FlatpakPermissionsModel = GObject.registerClass({
                     .split(';');
 
                 values.forEach(option => {
+                    /* Flatseal does not support conditionals, but skips them
+                     * to avoid corrupting the overrides file. */
                     if (option.startsWith(CONDITIONAL_PREFIX))
                         return;
 
--- a/src/models/permissions.js
+++ b/src/models/permissions.js
@@ -107,6 +107,7 @@ function generate() {
 }
 
 var DELAY = 500;
+var CONDITIONAL_PREFIX = 'if:';
 
 var FlatpakPermissionsModel = GObject.registerClass({
     GTypeName: 'FlatpakPermissionsModel',
@@ -189,6 +190,9 @@ var FlatpakPermissionsModel = GObject.registerClass({
                     .split(';');
 
                 values.forEach(option => {
+                    if (option.startsWith(CONDITIONAL_PREFIX))
+                        return;
+
                     model = this.constructor._find(`${group}_${key}_${option.replace('!', '')}`);
 
                     if (model === null)
--- a/tests/content/system/flatpak/app/com.test.Conditional/current/active/metadata
+++ b/tests/content/system/flatpak/app/com.test.Conditional/current/active/metadata
@@ -1,0 +1,8 @@
+[Application]
+name=com.test.Conditional
+runtime=org.gnome.Platform/x86_64/master
+sdk=org.gnome.Sdk/x86_64/master
+command=test
+
+[Context]
+sockets=x11;if:x11:!has-wayland;
--- a/tests/content/system/flatpak/app/com.test.Conditional/current/active/metadata
+++ b/tests/content/system/flatpak/app/com.test.Conditional/current/active/metadata
@@ -6,4 +6,3 @@ command=test
 
 [Context]
 sockets=x11;if:x11:!has-wayland;
-devices=all;
--- a/tests/content/system/flatpak/app/com.test.Conditional/current/active/metadata
+++ b/tests/content/system/flatpak/app/com.test.Conditional/current/active/metadata
@@ -6,3 +6,4 @@ command=test
 
 [Context]
 sockets=x11;if:x11:!has-wayland;
+devices=all;
--- a/tests/content/user/flatpak/overrides/com.test.Conditional
+++ b/tests/content/user/flatpak/overrides/com.test.Conditional
@@ -1,3 +1,2 @@
 [Context]
-sockets=!x11;if:x11:!has-wayland;
 devices=all;if:all:!has-input-device;
--- a/tests/content/user/flatpak/overrides/com.test.Conditional
+++ b/tests/content/user/flatpak/overrides/com.test.Conditional
@@ -1,0 +1,3 @@
+[Context]
+sockets=!x11;if:x11:!has-wayland;
+devices=all;if:all:!has-input-device;
--- a/tests/src/testModels.js
+++ b/tests/src/testModels.js
@@ -1472,8 +1472,9 @@ describe('Model', function() {
         GLib.timeout_add(GLib.PRIORITY_HIGH, delay + 1, () => {
             const group = permissionsDefault.constructor.getGroupForProperty('sockets-x11');
             expect(has(_conditionalOverride, group, 'sockets', '!x11')).toBe(true);
-            expect(has(_conditionalOverride, group, 'devices', '!all')).toBe(true);
-            expect(hasInTotal(_conditionalOverride)).toEqual(2);
+            expect(has(_conditionalOverride, group, 'sockets', 'if:x11:!has-wayland')).toBe(false);
+            expect(has(_conditionalOverride, group, 'devices', 'if:all:!has-input-device')).toBe(false);
+            expect(hasInTotal(_conditionalOverride)).toEqual(1);
             done();
             return GLib.SOURCE_REMOVE;
         });
--- a/tests/src/testModels.js
+++ b/tests/src/testModels.js
@@ -1467,12 +1467,13 @@ describe('Model', function() {
 
         GLib.setenv('FLATPAK_USER_DIR', _tmp, true);
         permissionsDefault.set_property('sockets-x11', false);
+        permissionsDefault.set_property('devices-all', false);
 
         GLib.timeout_add(GLib.PRIORITY_HIGH, delay + 1, () => {
             const group = permissionsDefault.constructor.getGroupForProperty('sockets-x11');
             expect(has(_conditionalOverride, group, 'sockets', '!x11')).toBe(true);
-            expect(has(_conditionalOverride, group, 'sockets', 'if:x11:!has-wayland')).toBe(false);
-            expect(has(_conditionalOverride, group, 'devices', 'if:all:!has-input-device')).toBe(false);
+            expect(has(_conditionalOverride, group, 'devices', '!all')).toBe(true);
+            expect(hasInTotal(_conditionalOverride)).toEqual(2);
             done();
             return GLib.SOURCE_REMOVE;
         });
--- a/tests/src/testModels.js
+++ b/tests/src/testModels.js
@@ -1462,33 +1462,16 @@ describe('Model', function() {
     });
 
     it('does not write conditional permissions back', function(done) {
-        GLib.setenv('FLATPAK_USER_DIR', _tmp, true);
+        GLib.setenv('FLATPAK_USER_DIR', _user, true);
         permissionsDefault.appId = _conditionalAppId;
 
+        GLib.setenv('FLATPAK_USER_DIR', _tmp, true);
         permissionsDefault.set_property('sockets-x11', false);
 
         GLib.timeout_add(GLib.PRIORITY_HIGH, delay + 1, () => {
             const group = permissionsDefault.constructor.getGroupForProperty('sockets-x11');
             expect(has(_conditionalOverride, group, 'sockets', '!x11')).toBe(true);
             expect(has(_conditionalOverride, group, 'sockets', 'if:x11:!has-wayland')).toBe(false);
-            done();
-            return GLib.SOURCE_REMOVE;
-        });
-
-        update();
-    });
-
-    it('does not write conditional overrides back', function(done) {
-        GLib.setenv('FLATPAK_USER_DIR', _user, true);
-        permissionsDefault.appId = _conditionalAppId;
-
-        GLib.setenv('FLATPAK_USER_DIR', _tmp, true);
-        expect(permissionsDefault.devices_all).toBe(true);
-        permissionsDefault.set_property('devices-all', false);
-
-        GLib.timeout_add(GLib.PRIORITY_HIGH, delay + 1, () => {
-            const group = permissionsDefault.constructor.getGroupForProperty('sockets-x11');
-            expect(has(_conditionalOverride, group, 'sockets', '!x11')).toBe(true);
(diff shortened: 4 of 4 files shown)


tests/content/system/flatpak/app/com.test.Conditional/current/active/metadata
```diff
@@ -0,0 +1,8 @@
+[Application]
+name=com.test.Conditional
+runtime=org.gnome.Platform/x86_64/master
+sdk=org.gnome.Sdk/x86_64/master
+command=test
+
+[Context]
+shared=network;
```
tchx84: Instead of non-conditional permission here, we should test an actual conditional permission, to cover the case where the app was built with such permission.

tests/content/user/flatpak/overrides/com.test.Conditional
```diff
@@ -0,0 +1,2 @@
+[Context]
+filesystems=if host-os=linux;home;
```
tchx84: Can you provide a `flatpak override` CLI example that would produce this?

tests/src/testModels.js
```diff
@@ -1452,6 +1455,31 @@ describe('Model', function() {
         update();
     });
 
+    it('ignores conditional permissions', function() {
```
tchx84, line 1458: I think this test should only focus on testing conditionals from the original app and not from the overrides.

tests/src/testModels.js
```diff
@@ -1452,6 +1455,31 @@ describe('Model', function() {
...
+        expect(permissionsDefault.filesystems_home).toBe(true);
+        expect(permissionsDefault.filesystems_other).toEqual('');
+    });
+
+    it('does not write conditional permissions back', function(done) {
+        GLib.setenv('FLATPAK_USER_DIR', _user, true);
+        permissionsDefault.appId = _conditionalAppId;
+
+        GLib.setenv('FLATPAK_USER_DIR', _tmp, true);
+        permissionsDefault.set_property('shared-network', false);
```
tchx84: What would be negating this shared-network permission? I suspect this would not pass.

src/models/permissions.js
```diff
@@ -189,6 +189,9 @@ var FlatpakPermissionsModel = GObject.registerClass({
                     .split(';');
 
                 values.forEach(option => {
+                    if (option.startsWith('if '))
```
tchx84: Can you check there aren't other possible syntax to express this? Also, it would be better to move this string to a global variable to make it more readable and prepare it in case we need to use it in some other place.

tchx84 (changes requested):

tests/content/user/flatpak/overrides/com.test.Conditional
Malika7188: Apologies, this was an incorrect syntax. I have removed this fixture file and replaced it with the correct conditional syntax in the app metadata instead. I have verified by testing with multiple conditional flags (-`-socket-if`, `--device-if`, `--share-if`, `--allow-if`) which confirmed all conditional permissions are stored with the if: prefix in the metadata i.e `flatpak build-finish --share-if=network:has-wayland`

src/models/permissions.js
Malika7188: I have added `CONDITIONAL_PREFIX = 'if:'` as a global variable, tested and pushed pushed the updated changes

tests/src/testModels.js
```diff
@@ -1452,6 +1455,29 @@ describe('Model', function() {
         update();
     });
 
+    it('ignores conditional permissions', function() {
+        permissionsDefault.appId = _conditionalAppId;
+
+        expect(permissionsDefault.sockets_wayland).toBe(true);
+    });
+
+    it('does not write conditional permissions back', function(done) {
```
tchx84, line 1464: I noticed you removed the override. We still want to check that, if someone uses `flatpak override` to add a conditional permission to an app, then uses Flatseal to change something on that same app, Flatseal would still ignore the conditional part and not write it back to the override file.

Please leave a comment in this PR with the `flatpak override` command to reproduce this test manually 🙏

tchx84 (changes requested):

tests/src/testModels.js
Malika7188: `flatpak override` rejects conditional syntax directly (e.g.` flatpak override --user --socket=if:wayland:has-wayland` returns "Unknown socket type"). 
To test the behavior, I manually added a conditional permission to the override file:
` printf '[Context]\nsockets=wayland;if:wayland:has-wayland;\n' > ~/.local/share/flatpak/overrides/<app-id>
` 
Then open Flatseal, modify any permission, and check the file `cat ~/.local/share/flatpak/overrides/<app`>. The conditional will be ignored and not written back.

tchx84: That's because the syntax you're using in that example is wrong, please read the Flatpak documentation carefully.

https://docs.flatpak.org/en/latest/sandbox-permissions.html#conditional-permissions

tests/content/user/flatpak/overrides/com.test.Conditional
```diff
@@ -0,0 +1,2 @@
+[Context]
+sockets=wayland;if:wayland:has-wayland;
```
tchx84: It's better if this uses a different conditional permission from the default app. Also, check the documentation I referenced and use the same examples from there (it's better to have semantically correct examples in the tests)

tests/src/testModels.js
Malika7188: I've updated the fixture files to use the exact examples from the Flatpak documentation.
  
The metadata file now uses `sockets=x11;if:x11:!has-wayland;` and the user override file uses a different conditional type ` devices=all;if:all:!has-input-device; `,  so the two fixtures test different permission categories.

To verify the correct syntax, I used:

`  flatpak override --user '--socket-if=x11:!has-wayland' <app-id>`

  Which produced:
```
  [Context]
  sockets=session-bus;x11;if:x11:!has-wayland;
```

I also tested the fix manually. I added a conditional permission to an app's override file using 
`flatpak override --user '--socket-if=x11:!has-wayland' <app-id>`, then edited the app in Flatseal. 

Before:
  `sockets=session-bus;x11;if:x11:!has-wayland;`
  After toggling X11 off in Flatseal:
  `sockets=session-bus;!x11`
  The conditional was correctly ignored and not written back. All three unit tests pass as well.

tests/content/user/flatpak/overrides/com.test.Conditional
```diff
@@ -0,0 +1,3 @@
+[Context]
+sockets=!x11;if:x11:!has-wayland;
```
tchx84: Not entirely sure I understand why having this `!x11:if...` negation is here. Specially since you set it to false in the test later, which seems redundant.

I think you can just remove this `!x11;if...`, keep setting it to false in the test, and then would make more sense.

tests/src/testModels.js
```diff
@@ -1452,6 +1455,48 @@ describe('Model', function() {
...
+            expect(has(_conditionalOverride, group, 'sockets', '!x11')).toBe(true);
+            expect(has(_conditionalOverride, group, 'sockets', 'if:x11:!has-wayland')).toBe(false);
+            done();
+            return GLib.SOURCE_REMOVE;
+        });
+
+        update();
+    });
+
+    it('does not write conditional overrides back', function(done) {
```
tchx84: Maybe merge both "not write back" tests together?

tests/src/testModels.js
```diff
@@ -1452,6 +1455,48 @@ describe('Model', function() {
...
+        GLib.setenv('FLATPAK_USER_DIR', _user, true);
+        permissionsDefault.appId = _conditionalAppId;
+
+        GLib.setenv('FLATPAK_USER_DIR', _tmp, true);
+        expect(permissionsDefault.devices_all).toBe(true);
+        permissionsDefault.set_property('devices-all', false);
+
+        GLib.timeout_add(GLib.PRIORITY_HIGH, delay + 1, () => {
+            const group = permissionsDefault.constructor.getGroupForProperty('sockets-x11');
+            expect(has(_conditionalOverride, group, 'sockets', '!x11')).toBe(true);
```
tchx84, line 1477: Why checking for x11 if this test is modifying `devices-all`?

tchx84 (changes requested):

tests/src/testModels.js
Malika7188: I have fixed the requested changes. I removed the redundant `sockets=!x11;if:x11:!has-wayland;` entry from the user override fixture and merged the two "not write back" tests into a single test. All tests still pass.

tests/src/testModels.js
```diff
@@ -1452,6 +1455,31 @@ describe('Model', function() {
...
+        expect(permissionsDefault.sockets_x11).toBe(true);
+    });
+
+    it('does not write conditional permissions back', function(done) {
+        GLib.setenv('FLATPAK_USER_DIR', _user, true);
+        permissionsDefault.appId = _conditionalAppId;
+
+        GLib.setenv('FLATPAK_USER_DIR', _tmp, true);
+        permissionsDefault.set_property('sockets-x11', false);
+
```
tchx84, line 1474: To make it easier to read and connect with the fixtures, I would also set devices-all to False here and check that `!all` is present.

tchx84 (changes requested):

tests/src/testModels.js
tchx84: Also, check for the `hasInTotal` helper. There are plenty of examples in existing tests.

Malika7188: I have added `devices-all` toggle alongside `sockets-x11` and used `hasInTotal` to verify the total count of values written to the override

tests/content/system/flatpak/app/com.test.Conditional/current/active/metadata
```diff
@@ -0,0 +1,9 @@
+[Application]
+name=com.test.Conditional
+runtime=org.gnome.Platform/x86_64/master
+sdk=org.gnome.Sdk/x86_64/master
+command=test
+
+[Context]
+sockets=x11;if:x11:!has-wayland;
+devices=all;
```
tchx84: Can you explain the reasoning for `devices=all` here? Having it on the overrides file should be enough.

tests/src/testModels.js
```diff
@@ -1452,6 +1455,32 @@ describe('Model', function() {
...
+        permissionsDefault.appId = _conditionalAppId;
+
+        GLib.setenv('FLATPAK_USER_DIR', _tmp, true);
+        permissionsDefault.set_property('sockets-x11', false);
+        permissionsDefault.set_property('devices-all', false);
+
+        GLib.timeout_add(GLib.PRIORITY_HIGH, delay + 1, () => {
+            const group = permissionsDefault.constructor.getGroupForProperty('sockets-x11');
+            expect(has(_conditionalOverride, group, 'sockets', '!x11')).toBe(true);
+            expect(has(_conditionalOverride, group, 'devices', '!all')).toBe(true);
```
tchx84: Even if it's redundant. keep the 2 extra checks for not having the conditional parts here. It makes the test easier to understand.

src/models/permissions.js
```diff
@@ -189,6 +190,9 @@ var FlatpakPermissionsModel = GObject.registerClass({
                     .split(';');
 
                 values.forEach(option => {
+                    if (option.startsWith(CONDITIONAL_PREFIX))
```
tchx84, line 195: Last thing, maybe add a comment on top of this line explaining that Flatseal still doesn't support conditionals but at least won't leave the overrides file in a broken state when there are conditionals.

tchx84 (changes requested):

tests/content/system/flatpak/app/com.test.Conditional/current/active/metadata
Malika7188: `devices=all;` is needed in the metadata because Flatseal uses it to decide whether a change should be written to the override file. When `devices-all` is toggled off, Flatseal checks if all was originally present in the metadata. If it wasn’t, it treats the value as the default and does not write anything, so `!all `would not appear in the override file and the test would fail

Adding d`evices=all;` ensures the toggle produces a real change, so` !all `is correctly written to the override file

tchx84: That's not quite correct. In this particular case, the override file *is* providing the `devices=all` permission that we will later set to false. What this shows is that the current "not write back" test still needs tweaking.

Remove this "device=all" permission here, keep the override file as is, and the result override file (after the test) will only include "!x11" and count should be 1.

tchx84: Try this experiment manually:

1. Run `flatpak --user override --socket=cups org.gnome.Showtime`.
2. Open Flatseal, you will see that Showtime does show the "cups" toggle, and you can toggle it off.
3. After you toggle it off, there will be no override file anymore.

NOTE that Showtime does not have cups originally.

Malika7188: Oh yeah, you're right. That makes sense. Let me fix it

src/models/permissions.js
Malika7188: I have removed `devices=all;` from the metadata, updated the test checks, and added the comment in permissions.js.

tests/src/testModels.js
```diff
@@ -1452,6 +1455,33 @@ describe('Model', function() {
...
 
+    it('ignores conditional permissions', function() {
+        permissionsDefault.appId = _conditionalAppId;
+
+        expect(permissionsDefault.sockets_x11).toBe(true);
+    });
+
+    it('does not write conditional permissions back', function(done) {
+        GLib.setenv('FLATPAK_USER_DIR', _user, true);
+        permissionsDefault.appId = _conditionalAppId;
```
tchx84, line 1466: Before it switches to tmp, let's actually verify that both `x11` and `all` are indeed enabled via original permissions and overrides.. To make this test more robust.

tchx84 (changes requested):

tests/src/testModels.js
Malika7188: I've implemented this. The tests now verifies that both `x11` and `all` are enabled beofore switching to `_tmp`.

PR #880, opened 2026-06-17
fix: USB portal support fixes and test coverage

Malika7188:
Expose the portal's enumerable-devices (`--usb`) and hidden-devices (`--nousb`) overrides as two independent properties, `usb` and `usb-hidden`. Each tracks its own originals, globals, and overrides, and neither depends on the other's state when processing changes, allowing a device to exist in both the allowed and blocked lists without either property modifying the other.

Add a USB permissions row with a USB-specific validation regex, a status icon indicating whether an entry is original, global, or user-added, and a remove button that is only enabled for user-added entries.

Handle empty-string values safely when loading and updating USB portal permissions.

Extend test coverage for model loading and saving, widget validation, and remove-button sensitivity.

### DIFF
--- a/src/models/permissions.js
+++ b/src/models/permissions.js
@@ -29,6 +29,7 @@ const {FlatpakFilesystemsOtherModel} = imports.models.filesystemsOther;
 const {FlatpakVariablesModel} = imports.models.variables;
 const {FlatpakSessionBusModel} = imports.models.sessionBus;
 const {FlatpakSystemBusModel} = imports.models.systemBus;
+const {FlatpakUsbModel} = imports.models.usb;
 const {FlatsealOverrideStatus} = imports.models.overrideStatus;
 const {isGlobalOverride} = imports.models.globalModel;
 const {applications, filesystems, persistent, portals} = imports.models;
@@ -46,6 +47,7 @@ const MODELS = {
     variables: new FlatpakVariablesModel(),
     system: new FlatpakSystemBusModel(),
     session: new FlatpakSessionBusModel(),
+    usb: new FlatpakUsbModel(),
     portals: portals.getDefault(),
     unsupported: new FlatpakUnsupportedModel(),
 };
--- a/src/models/usb.js
+++ b/src/models/usb.js
@@ -1,0 +1,118 @@
+/* exported FlatpakUsbModel */
+
+/* usb.js
+ *
+ * Copyright 2025 Martin Abente Lahaye
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const {GObject} = imports.gi;
+
+const {FlatpakSharedModel} = imports.models.shared;
+const {FlatsealOverrideStatus} = imports.models.overrideStatus;
+
+var FlatpakUsbModel = GObject.registerClass({
+    GTypeName: 'FlatpakUsbModel',
+}, class FlatpakUsbModel extends FlatpakSharedModel {
+    _init() {
+        super._init({});
+        this.reset();
+    }
+
+    getPermissions() {
+        return {
+            usb: {
+                supported: this._info.supports('1.15.11'),
+                description: _('Devices'),
+                option: null,
+                value: this.constructor.getDefault(),
+                example: _('e.g. vnd:0123+dev:4567'),
+            },
+        };
+    }
+
+    static getDefault() {
+        return '';
+    }
+
+    static getType() {
+        return 'usb';
+    }
+
+    static getGroup() {
+        return 'USB Devices';
+    }
+
+    static getKey() {
+        return 'enumerable-devices';
+    }
+
+    static getStyle() {
+        return 'usb';
+    }
+
+    static getTitle() {
+        return 'USB';
+    }
+
+    static getDescription() {
+        return _('List of devices matching the query visible to the USB portal');
+    }
+
+    getOptions() { // eslint-disable-line class-methods-use-this
+        return null;
+    }
+
+    static isNegated(value) {
+        return value.startsWith('!');
+    }
+
+    static negate(value) {
+        if (this.isNegated(value))
+            return value.replace('!', '');
+        return `!${value}`;
+    }
+
+    updateFromProxyProperty(property, value) {
+        const values = new Set(this.constructor.deserialize(value));
+
+        const baseline = this._originals.union(this._globals);
+        const added = values.difference(baseline);
+        const removed = new Set([...baseline.difference(values)]
+            .map(d => this.constructor.negate(d)));
+
+        this._overrides = added.union(removed);
+    }
+
+    updateStatusProperty(proxy) {
+        const statuses = this.constructor.deserialize(proxy.usb)
+            .map(d => this._getStatusForPermission(d));
+
+        proxy.set_property('usb-status', this.constructor.serialize(statuses));
+    }
+
+    updateProxyProperty(proxy) {
+        const originals = [...this._originals]
+            .filter(o => !this._globals.has(this.constructor.negate(o)))
+            .filter(o => !this._overrides.has(this.constructor.negate(o)));
+
+        const globals = [...this._globals]
+            .filter(g => !this._overrides.has(this.constructor.negate(g)));
+
+        const usb = [...originals, ...globals, ...this._overrides];
+
+        proxy.set_property('usb', this.constructor.serialize(usb));
+    }
+});
--- a/src/models/usb.js
+++ b/src/models/usb.js
@@ -3,6 +3,7 @@
 /* usb.js
  *
  * Copyright 2025 Martin Abente Lahaye
+ * Copyright 2026 Malika Odeny Asman
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -35,7 +36,14 @@ var FlatpakUsbModel = GObject.registerClass({
         return {
             usb: {
                 supported: this._info.supports('1.15.11'),
-                description: _('Devices'),
+                description: _('Allowed devices'),
+                option: null,
+                value: this.constructor.getDefault(),
+                example: _('e.g. vnd:0123+dev:4567'),
+            },
+            'usb-hidden': {
+                supported: this._info.supports('1.15.11'),
+                description: _('Blocked devices'),
                 option: null,
                 value: this.constructor.getDefault(),
                 example: _('e.g. vnd:0123+dev:4567'),
@@ -56,7 +64,7 @@ var FlatpakUsbModel = GObject.registerClass({
     }
 
     static getKey() {
-        return 'enumerable-devices';
+        return null;
     }
 
     static getStyle() {
@@ -68,51 +76,104 @@ var FlatpakUsbModel = GObject.registerClass({
     }
 
     static getDescription() {
-        return _('List of devices matching the query visible to the USB portal');
+        return _('List of USB devices accessible through the portal');
     }
 
     getOptions() { // eslint-disable-line class-methods-use-this
         return null;
     }
 
-    static isNegated(value) {
-        return value.startsWith('!');
+    reset() {
+        super.reset();
+        this._hiddenOriginals = new Set();
+        this._hiddenGlobals = new Set();
+        this._hiddenOverrides = new Set();
     }
 
-    static negate(value) {
-        if (this.isNegated(value))
-            return value.replace('!', '');
-        return `!${value}`;
+    _findProperHiddenSet(overrides, global) {
+        if (overrides && global)
+            return this._hiddenGlobals;
+        if (overrides && !global)
+            return this._hiddenOverrides;
+        return this._hiddenOriginals;
+    }
+
+    _getHiddenStatusForDevice(device) {
+        let status = FlatsealOverrideStatus.ORIGINAL;
+        if (this._hiddenGlobals.has(device))
+            status = FlatsealOverrideStatus.GLOBAL;
+        if (this._hiddenOverrides.has(device))
+            status = FlatsealOverrideStatus.USER;
+        return status;
     }
 
     updateFromProxyProperty(property, value) {
-        const values = new Set(this.constructor.deserialize(value));
+        const devices = new Set(this.constructor.deserialize(value)
+            .filter(d => d.length !== 0));
 
-        const baseline = this._originals.union(this._globals);
-        const added = values.difference(baseline);
-        const removed = new Set([...baseline.difference(values)]
-            .map(d => this.constructor.negate(d)));
-
-        this._overrides = added.union(removed);
+        if (property === 'usb') {
+            this._overrides = new Set([...devices]
+                .filter(d => !this._originals.has(d))
+                .filter(d => !this._globals.has(d)));
+        } else if (property === 'usb-hidden') {
+            this._hiddenOverrides = new Set([...devices]
+                .filter(d => !this._hiddenOriginals.has(d))
+                .filter(d => !this._hiddenGlobals.has(d)));
+        }
     }
 
     updateStatusProperty(proxy) {
-        const statuses = this.constructor.deserialize(proxy.usb)
+        const usbStatuses = this.constructor.deserialize(proxy.usb)
+            .filter(d => d.length !== 0)
             .map(d => this._getStatusForPermission(d));
+        proxy.set_property('usb-status', this.constructor.serialize(usbStatuses));
 
-        proxy.set_property('usb-status', this.constructor.serialize(statuses));
+        const hiddenStatuses = this.constructor.deserialize(proxy.usb_hidden)
+            .filter(d => d.length !== 0)
+            .map(d => this._getHiddenStatusForDevice(d));
+        proxy.set_property('usb-hidden-status', this.constructor.serialize(hiddenStatuses));
     }
 
     updateProxyProperty(proxy) {
-        const originals = [...this._originals]
-            .filter(o => !this._globals.has(this.constructor.negate(o)))
-            .filter(o => !this._overrides.has(this.constructor.negate(o)));
+        const allowed = new Set([...this._originals, ...this._globals, ...this._overrides]);
+        const blocked = new Set([...this._hiddenOriginals, ...this._hiddenGlobals, ...this._hiddenOverrides]);
 
-        const globals = [...this._globals]
-            .filter(g => !this._overrides.has(this.constructor.negate(g)));
+        proxy.set_property('usb', this.constructor.serialize([...allowed]));
+        proxy.set_property('usb-hidden', this.constructor.serialize([...blocked]));
+    }
 
-        const usb = [...originals, ...globals, ...this._overrides];
+    loadFromKeyFile(group, key, value, overrides, global) {
+        const devices = this.constructor.deserialize(value)
+            .filter(d => d.length !== 0);
 
-        proxy.set_property('usb', this.constructor.serialize(usb));
+        if (key === 'enumerable-devices') {
+            const set = this._findProperSet(overrides, global);
+            devices.forEach(d => set.add(d));
+        } else if (key === 'hidden-devices') {
+            const set = this._findProperHiddenSet(overrides, global);
+            devices.forEach(d => set.add(d));
+        }
+    }
+
+    saveToKeyFile(keyFile) {
+        const group = this.constructor.getGroup();
+
+        this._overrides.forEach(value => {
+            try {
+                const existing = keyFile.get_value(group, 'enumerable-devices');
+                keyFile.set_value(group, 'enumerable-devices', `${value};${existing}`);
+            } catch (err) {
+                keyFile.set_value(group, 'enumerable-devices', `${value}`);
+            }
+        });
+
+        this._hiddenOverrides.forEach(value => {
+            try {
+                const existing = keyFile.get_value(group, 'hidden-devices');
+                keyFile.set_value(group, 'hidden-devices', `${value};${existing}`);
+            } catch (err) {
+                keyFile.set_value(group, 'hidden-devices', `${value}`);
+            }
+        });
     }
 });
--- a/src/widgets/usbRow.js
+++ b/src/widgets/usbRow.js
@@ -1,0 +1,122 @@
+/* exported FlatsealUsbRow validity */
+
+/* usbRow.js
+ *
+ * Copyright 2025 Martin Abente Lahaye
+ * Copyright 2026 Malika Odeny Asman
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+const {GObject, Gtk} = imports.gi;
+
+const {FlatsealOverrideStatusIcon} = imports.widgets.overrideStatusIcon;
+const {FlatsealOverrideStatus} = imports.models.overrideStatus;
+
+const _propFlags = GObject.ParamFlags.READWRITE | GObject.ParamFlags.CONSTRUCT;
+
+/* https://github.com/flatpak/flatpak/blob/main/doc/flatpak-build-finish.xml */
+const _deviceExp = '(vnd:[0-9a-fA-F]{4}|dev:[0-9a-fA-F]{4}|cls:[0-9a-fA-F]{2}:(?:[0-9a-fA-F]{2}|\\*))';
+const EXP = new RegExp(`^(?!(?=.*dev:)(?!.*vnd:))(all|${_deviceExp}(\\+${_deviceExp})*)$`);
+
+var validity = {
+    VALID: 'valid',
+    NOTVALID: 'not-valid',
+};
+
+const _notValidMsg = _('This is not a valid option');
+
+
+var FlatsealUsbRow = GObject.registerClass({
+    GTypeName: 'FlatsealUsbRow',
+    Template: 'resource:///com/github/tchx84/Flatseal/widgets/usbRow.ui',
+    InternalChildren: ['entry', 'button', 'image', 'statusBox'],
+    Properties: {
+        text: GObject.ParamSpec.string(
+            'text',
+            'text',
+            'text',
+            _propFlags, ''),
+    },
+    Signals: {
+        'remove-requested': {},
+    },
+}, class FlatsealUsbRow extends Gtk.Box {
+    _init() {
+        super._init({});
+        this._setup();
+    }
+
+    _setup() {
+        this._expression = new RegExp(EXP);
+
+        this._entry.connect('notify::text', this._changed.bind(this));
+        this._button.connect('clicked', this._remove.bind(this));
+
+        this._statusIcon = new FlatsealOverrideStatusIcon();
+        this._statusBox.append(this._statusIcon);
+
+        this._validate();
+    }
+
+    _remove() {
+        this.emit('remove-requested');
+    }
+
+    _changed() {
+        this._validate();
+        this.notify('text');
+    }
+
+    _validate() {
+        const context = this.get_style_context();
+
+        if (context.has_class(validity.VALID))
+            context.remove_class(validity.VALID);
+        else if (context.has_class(validity.NOTVALID))
+            context.remove_class(validity.NOTVALID);
+
+        if (this._expression.test(this.text)) {
+            context.add_class(validity.VALID);
+            this._image.set_tooltip_text('');
+        } else {
+            context.add_class(validity.NOTVALID);
+            this._image.set_tooltip_text(_notValidMsg);
+        }
+    }
+
+    get text() {
+        if (!this._entry)
+            return '';
+        return this._entry.get_text();
+    }
+
+    set text(text) {
+        if (this.text === text)
+            return;
+        this._entry.set_text(text);
+    }
+
+    get status() {
+        return this._statusIcon.status;
+    }
+
+    set status(status) {
+        this._statusIcon.status = status;
+
+        /* only user-added entries can be edited or removed; original and
+         * global entries have no override to change and would have no effect */
+        this.sensitive = status === FlatsealOverrideStatus.USER;
+    }
+});
--- a/tests/content/statuses/flatpak/overrides/com.test.Usb
+++ b/tests/content/statuses/flatpak/overrides/com.test.Usb
@@ -1,0 +1,3 @@
+[USB Devices]
+enumerable-devices=vnd:0789;
+hidden-devices=vnd:0111;
--- a/tests/content/user/flatpak/overrides/com.test.Usb
+++ b/tests/content/user/flatpak/overrides/com.test.Usb
@@ -1,0 +1,3 @@
+[USB Devices]
+enumerable-devices=vnd:0456;
+hidden-devices=vnd:0789;
--- a/tests/src/testModels.js
+++ b/tests/src/testModels.js
@@ -33,7 +33,7 @@ const {
 
 setup();
 
-const _totalPermissions = 41;
+const _totalPermissions = 43;
 
 const _basicAppId = 'com.test.Basic';
 const _basicNegatedAppId = 'com.test.BasicNegated';
@@ -47,6 +47,7 @@ const _overridenAppId = 'com.test.Overriden';
 const _extraAppId = 'com.test.Extra';
 const _environmentAppId = 'com.test.Environment';
 const _busAppId = 'com.test.Bus';
+const _usbAppId = 'com.test.Usb';
 const _variablesAppId = 'com.test.Variables';
 const _trailingSemicolonId = 'com.test.TrailingSemicolon';
 const _filesystemWithMode = 'com.test.FilesystemWithMode';
@@ -81,6 +82,7 @@ const _unsupportedOverride = GLib.build_filenamev([_overrides, _unsupportedAppId
 const _overridenOverride = GLib.build_filenamev([_overrides, _overridenAppId]);
 const _environmentOverride = GLib.build_filenamev([_overrides, _environmentAppId]);
 const _busOverride = GLib.build_filenamev([_overrides, _busAppId]);
+const _usbOverride = GLib.build_filenamev([_overrides, _usbAppId]);
 const _filesystemWithModeOverride = GLib.build_filenamev([_overrides, _filesystemWithMode]);
 const _resetModeOverride = GLib.build_filenamev([_overrides, _resetModeId]);
 const _conditionalOverride = GLib.build_filenamev([_overrides, _conditionalAppId]);
@@ -88,6 +90,9 @@ const _globalWithGlobalOverride = GLib.build_filenamev([_global, 'overrides', _g
 
 const _sessionGroup = 'Session Bus Policy';
 const _key = 'filesystems';
+const _usbGroup = 'USB Devices';
+const _usbKey = 'enumerable-devices';
+const _usbHiddenKey = 'hidden-devices';
 
 const _flatpakConfig = GLib.build_filenamev(['..', 'tests', 'content']);
 
@@ -134,6 +139,7 @@ describe('Model', function() {
         GLib.unlink(_unsupportedOverride);
         GLib.unlink(_environmentOverride);
         GLib.unlink(_busOverride);
+        GLib.unlink(_usbOverride);
         GLib.unlink(_filesystemWithModeOverride);
         GLib.unlink(_resetModeOverride);
         GLib.unlink(_conditionalOverride);
@@ -718,7 +724,7 @@ describe('Model', function() {
 
         const total = permissionsDefault.getAll().filter(p => p.supported).length;
 
-        expect(total).toEqual(_totalPermissions - 11);
+        expect(total).toEqual(_totalPermissions - 13);
     });
 
     it('handles missing .flatpak-info', function() {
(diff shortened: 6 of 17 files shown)


tests/src/testModels.js
```diff
@@ -1493,4 +1501,98 @@ describe('Model', function() {
...
+    it('removes USB enumerable device', function(done) {
+        GLib.setenv('FLATPAK_USER_DIR', _tmp, true);
+        permissionsDefault.appId = _usbAppId;
+
+        expect(permissionsDefault.usb).toEqual('vnd:0123');
+
+        permissionsDefault.set_property('usb', '');
+
+        GLib.timeout_add(GLib.PRIORITY_HIGH, delay + 1, () => {
+            expect(has(_usbOverride, _usbGroup, _usbKey, '!vnd:0123')).toBe(true);
```
tchx84: Can you show an example, using `flatpak override` that generates this?

Malika7188: Apologies, there is no Flatpak command that produces `!vnd:0123.` USB permissions are represented through enumerable-devices and hidden-devices only, there is no negated entries. I've corrected the implementation and updated the PR.

tchx84: If that's the case, given the current implementation, what would happen if there's an app with usb device permission in the originals, but then it's removed using the UI?

Malika7188: The device from originals will persist regardless of what the user does in the UI, since Flatpak provides no way to negate `enumerable-devices`. This is a Flatpak limitation and I think the best way to handle this is by adding a test to document this behavior. What do you think?

tchx84: I believe you need to look at this in more depth.

There's a reason why `flatpak override` exposes `--usb` and `--nousb`, and the Flatseal UI should model this in a way that provides a good UX.

src/models/permissions.js
```diff
@@ -130,6 +129,7 @@ var FlatpakPermissionsModel = GObject.registerClass({
         this._monitors = [];
         this._monitorsDelayedHandlerId = 0;
         this._changesByUser = 0;
+        this._pendingProperties = new Set();
```
tchx84: This seems unrelated, what's this?

src/models/usb.js
```diff
@@ -0,0 +1,201 @@
+/* exported FlatpakUsbModel */
+
+/* usb.js
+ *
+ * Copyright 2025 Martin Abente Lahaye
```
tchx84, line 5: You should include a line with your copyright line as well.

src/widgets/usbRow.js
```diff
@@ -0,0 +1,115 @@
+/* exported FlatsealUsbRow */
+
+/* usbRow.js
+ *
+ * Copyright 2025 Martin Abente Lahaye
```
tchx84, line 5: Same here, since you expanded this significantly.

tests/content/statuses/flatpak/overrides/com.test.Usb
```diff
@@ -0,0 +1,2 @@
+[USB Devices]
+enumerable-devices=vnd:0789;
```
tchx84, line 2: We should also test loading a hidden device here? Since the previous separate file was removed.

tests/content/user/flatpak/overrides/com.test.Usb
```diff
@@ -0,0 +1,2 @@
+[USB Devices]
+enumerable-devices=vnd:0456;
```
tchx84, line 2: Same here, we should test loading `hidden-devices`.

tests/src/testModels.js
```diff
@@ -1493,4 +1499,131 @@ describe('Model', function() {
...
+        expect(permissionsDefault.usb).toEqual('vnd:0123');
+        expect(permissionsDefault.usb_hidden).toEqual('vnd:0408+dev:5348');
+
+        GLib.setenv('FLATPAK_USER_DIR', _user, true);
+        permissionsDefault.appId = _usbAppId;
+        expect(permissionsDefault.usb).toEqual('vnd:0123;vnd:0456');
+
+        GLib.setenv('FLATPAK_USER_DIR', _global, true);
+        permissionsDefault.appId = _usbAppId;
+        expect(permissionsDefault.usb).toEqual('vnd:0123;vnd:0456');
```
tchx84: Use a different usb device value for global to more clearly differentiate this case? Compared to the earlier "_user" case.

tests/src/testModels.js
```diff
@@ -1493,4 +1499,131 @@ describe('Model', function() {
...
+
+        GLib.setenv('FLATPAK_USER_DIR', _statuses, true);
+        permissionsDefault.appId = _usbAppId;
+        expect(permissionsDefault.usb_status).toEqual('original;global;user');
```
tchx84, line 1520: Needs same thing for hidden.

tests/src/testModels.js
```diff
@@ -1493,4 +1499,131 @@ describe('Model', function() {
...
+
+    it('adds USB devices', function(done) {
+        GLib.setenv('FLATPAK_USER_DIR', _tmp, true);
+        permissionsDefault.appId = _usbAppId;
+
+        expect(permissionsDefault.usb).toEqual('vnd:0123');
+        expect(permissionsDefault.usb_hidden).toEqual('vnd:0408+dev:5348');
+
+        permissionsDefault.set_property('usb', 'vnd:0123;vnd:0456');
+        permissionsDefault.set_property('usb-hidden', 'vnd:0408+dev:5348;vnd:0123');
```
tchx84: There's already a separate test to test the "switch from allowed to logic", so maybe use a different value (not `vnd:0123`), to make it clear that you're just testing adding something new to the usb-hidden prop and that it's getting written correctly.

tests/src/testModels.js
```diff
@@ -1493,4 +1499,131 @@ describe('Model', function() {
...
+            expect(has(_usbOverride, _usbGroup, _usbHiddenKey, 'vnd:0123')).toBe(true);
+            expect(has(_usbOverride, _usbGroup, _usbHiddenKey, 'vnd:0408+dev:5348')).toBe(false);
+            done();
+            return GLib.SOURCE_REMOVE;
+        });
+
+        update();
+    });
+
+    it('does not write override when USB devices unchanged', function(done) {
```
tchx84: I am not sure how useful this test is.

tests/src/testModels.js
```diff
@@ -1493,4 +1499,131 @@ describe('Model', function() {
...
+        GLib.setenv('FLATPAK_USER_DIR', _tmp, true);
+        permissionsDefault.appId = _usbAppId;
+
+        expect(permissionsDefault.usb).toEqual('vnd:0123');
+
+        permissionsDefault.set_property('usb', '');
+
+        GLib.timeout_add(GLib.PRIORITY_HIGH, delay + 1, () => {
+            expect(has(_usbOverride, _usbGroup, _usbHiddenKey, 'vnd:0123')).toBe(true);
+            expect(has(_usbOverride, _usbGroup, _usbKey, 'vnd:0123')).toBe(false);
```
tchx84: include the hasInTotal to double check everything's correct here.

tests/src/testModels.js
```diff
@@ -1493,4 +1499,131 @@ describe('Model', function() {
...
+        });
+
+        update();
+    });
+
+    it('restores original USB device to allowed when removed from blocked', function(done) {
+        GLib.setenv('FLATPAK_USER_DIR', _tmp, true);
+        permissionsDefault.appId = _usbAppId;
+
+        permissionsDefault.set_property('usb', '');
```
tchx84, line 1552: Before resetting this, you should confirm your assumptions by checking existing values.

tests/src/testModels.js
```diff
@@ -1493,4 +1499,131 @@ describe('Model', function() {
...
+        permissionsDefault.set_property('usb-hidden', 'vnd:0408+dev:5348;vnd:0123');
+
+        permissionsDefault.set_property('usb-hidden', 'vnd:0408+dev:5348');
```
tchx84: Why setting usb-hidden two times consecutively?

tests/src/testModels.js
```diff
@@ -1493,4 +1499,131 @@ describe('Model', function() {
...
+        });
+
+        update();
+    });
+
+    it('removes user-added USB device from allowed without blocking it', function(done) {
+        GLib.setenv('FLATPAK_USER_DIR', _tmp, true);
+        permissionsDefault.appId = _usbAppId;
+
+        permissionsDefault.set_property('usb', 'vnd:0123;vnd:0456');
```
tchx84: Please apply the same suggestions I made for the previous tests.

tests/src/testModels.js
```diff
@@ -1493,4 +1499,131 @@ describe('Model', function() {
...
+
+        update();
+    });
+
+    it('removes user-added USB device from blocked without allowing it', function(done) {
+        GLib.setenv('FLATPAK_USER_DIR', _tmp, true);
+        permissionsDefault.appId = _usbAppId;
+
+        permissionsDefault.set_property('usb-hidden', 'vnd:0408+dev:5348;vnd:0456');
+        permissionsDefault.set_property('usb-hidden', 'vnd:0408+dev:5348');
```
tchx84: Setting the same properly two times consecutively shouldn't make any difference, can you explain?

tchx84 (changes requested):

src/models/permissions.js
Malika7188: The `_pendingProperties` change is required for the unified USB model to work correctly.

Previously,` _updateModels` re-read every proxy property on every update, including properties the user didn't change. For most models this is fine, but `FlatpakUsbModel` has two properties that interact: changing usb automatically  computes new entries for `_hiddenOverrides`. The problem is that processing usb-hidden immediately after `resets _hiddenOverrides` from scratch using the stale proxy value, wiping out what was just computed.

` _pendingProperties` tracks which property the user actually changed so `_updateModels` only processes those. If the user only changed usb, `usb-hidden` is skipped, preserving the computed _hiddenOverrides. Then `_updateProperties()` syncs the  proxy to reflect the new model state.

Without this, removing an original device from the allowed list would silently leave it accessible, it would disappear from the UI but never actually get blocked.

tchx84: > The problem is that processing usb-hidden immediately after resets _hiddenOverrides from scratch using the stale proxy value, wiping out what was just computed.

This sounds more like an issue that could be addressed internally in the model, given that the behavior is predictable, and not worked around here.

Can provide step-by-step instructions to reproduce this scenario manually?

tchx84 (changes requested):
A thing we're missing here, since this is introducing a new row with regex validation, is a new test for it.

See `tests/src/testPathRow.js` for an example.

src/models/permissions.js
Malika7188: To reproduce manually: open Flatseal, `select com.test.Usb`, remove `vnd:0123` from Devices, without the fix it silently disappears instead of moving to Blocked devices.

Why this happens:

**Setup:** app metadata has `vnd:0123` in `enumerable-devices` and `vnd:0408+dev:5348` in hidden-devices.

**Step 1:** User opens Flatseal, selects the app. Model loads:
  - _originals = `{vnd:0123}`, _hiddenOriginals = `{vnd:0408+dev:5348}`
  - Proxy shows: usb = `'vnd:0123'`, usb-hidden = 'vnd:0408+dev:5348'

**Step 2**: User removes vnd:0123 from the allowed list. Proxy becomes:
  - usb = ' '
  - usb-hidden = `vnd:0408+dev:5348` ← user never touched this, still old value

**Step 3 :** 500ms later _updateModels fires. Processes usb first:
  - vnd:0123 missing from allowed → _hiddenOverrides.add(`vnd:0123`) ✓

**Step 4:** Same `_updateModels` call, processes usb-hidden next using the stale proxy value `vnd:0408+dev:5348`:
  - Wipes _hiddenOverrides completely → `_hiddenOverrides` = {}..gone

**Step 5:** Override file is saved with nothing in it. File gets deleted.

**Step 6:** Proxy syncs: `vnd:0123` reappears in the allowed list. Nothing was blocked.

 **Fix:** `_pendingProperties` tracks which property the user actually changed. `_updateModels` only re-reads those, so when only usb changed, usb-hidden is skipped and _hiddenOverrides computed in Step 3 is preserved when the file is saved.

- I've added a short video demonstrating the fix


https://github.com/user-attachments/assets/ac861a41-7e26-4f81-b54d-4bcea0c5ad2a

Malika7188:
> A thing we're missing here, since this is introducing a new row with regex validation, is a new test for it.
> 
> See `tests/src/testPathRow.js` for an example.

I've added `tests/src/testUsbRow.js` covering valid and invalid cases. All tests passed successfully.

src/models/permissions.js
tchx84: Thanks for the explanation, unfortunately, this shows that this approach still needs rethinking.

As I mentioned before, the top permission model shouldn't be working around an issue of this particular model. The issue here is that you have two properties on the same model and these properties interact (**plus, this workaround could have other side effects for the other models that can rely on all other models being updated**).

In fact, this is not a new scenario for Flatseal. E.g., see `filesystems` and `filesystemsOther` models for a somewhat similar scenario.

Therefore, the `usb` model should be able to address this without this workaround. To do this, the model can take advantage of the fact that these properties will always update in that particular order. The model should be aware of this fact and do the right thing. I would first exhaust this approach.

If it's not possible, then we should consider having just having 1 properly for both allowed/hidden, and have a bit more complex UI (**but I don't think we will need this**).

tests/src/testUsbRow.js
```diff
@@ -0,0 +1,94 @@
...
+            row.text = device;
+            const context = row.get_style_context();
+
+            expect(context.has_class(validity.VALID)).toBe(true);
+            expect(context.has_class(validity.NOTVALID)).toBe(false);
+        });
+    }
+
+    _handles('all devices', 'all');
+    _handles('negated all devices', '!all');
```
tchx84: Can you point me to the flatpak documentation that mentions this syntax?

tchx84 (changes requested):

tests/src/testUsbRow.js
Malika7188: Hi @tchx84  Sorry I missed this. I looked into the Flatpak documentation in detail and verified the correct USB query syntax directly against `flatpak override`. I've updated the USB query validation to match the supported syntax and corrected the tests accordingly.

tchx84: sure, but can you point me to the documentation? or source you used to construct that regex and tests?

I still see entries with the `!` prefix, btw.

Malika7188: I’ve updated the tests accordingly.

Here are the eferences I used:

- https://github.com/flatpak/flatpak/blob/main/doc/flatpak-build-finish.xml
- https://notes.igalia.com/DY0n8yK_TtyLtx_VJHC9AA?both

The Flatpak documentation was the primary source for the supported USB query syntax, and the previous notes helped while refining the implementation and tests.

Regarding the `!` cases: The `!` entries in the tests are now in `_catches`, which verifies the regex rejects them. they are invalid inputs. The `!` prefix is not part of the `--usb` syntax defined in `flatpak-build-finish.xml`.

src/models/permissions.js
Malika7188: Thanks for the feedback. I reworked the implementation so the fix is now entirely inside `FlatpakUsbModel` and removed `_pendingProperties` from `permissions.js` so no other model is affected by it

The model now skips reprocessing a property only when the incoming proxy value matches the last value it wrote back. This prevents an unchanged `usb-hidden` value from overwriting the auto-block computed from `usb` in the same update cycle, while still processing real user changes.

I've verified it with the full test suite, manual testing, and updated the USB tests to cover this scenario . Could you take another look?

tests/src/testModels.js
```diff
@@ -1493,4 +1499,166 @@ describe('Model', function() {
...
+        expect(permissionsDefault.usb).toEqual('vnd:0123;vnd:0456');
+        expect(permissionsDefault.usb_hidden).toContain('vnd:0789');
+
+        GLib.setenv('FLATPAK_USER_DIR', _global, true);
+        permissionsDefault.appId = _usbAppId;
+        expect(permissionsDefault.usb).toEqual('vnd:0123;vnd:0567');
+
+        GLib.setenv('FLATPAK_USER_DIR', _statuses, true);
+        permissionsDefault.appId = _usbAppId;
+        expect(permissionsDefault.usb_status).toEqual('original;global;user');
```
tchx84, line 1520: This particular test will be easier to read if you include an `expect` for the actual values these statuses correspond to.

tests/src/testModels.js
```diff
@@ -1493,4 +1499,166 @@ describe('Model', function() {
...
+        expect(permissionsDefault.usb_hidden).toEqual('vnd:0408+dev:5348');
+
+        permissionsDefault.set_property('usb', 'vnd:0123;vnd:0456');
+        permissionsDefault.set_property('usb-hidden', 'vnd:0408+dev:5348;vnd:0999');
+
+        GLib.timeout_add(GLib.PRIORITY_HIGH, delay + 1, () => {
+            expect(has(_usbOverride, _usbGroup, _usbKey, 'vnd:0456')).toBe(true);
+            expect(has(_usbOverride, _usbGroup, _usbKey, 'vnd:0123')).toBe(false);
+            expect(has(_usbOverride, _usbGroup, _usbHiddenKey, 'vnd:0999')).toBe(true);
+            expect(has(_usbOverride, _usbGroup, _usbHiddenKey, 'vnd:0408+dev:5348')).toBe(false);
```
tchx84, line 1539: Please follow my previous request of counting the overrides to all the new tests you're adding.

tests/src/testModels.js
```diff
@@ -1493,4 +1499,166 @@ describe('Model', function() {
...
+            expect(has(_usbOverride, _usbGroup, _usbHiddenKey, 'vnd:0123')).toBe(true);
+            expect(permissionsDefault.usb).not.toContain('vnd:0123');
+            done();
+            return GLib.SOURCE_REMOVE;
+        });
+
+        update();
+    });
+
+    it('restores auto-blocked original USB device to allowed when removed from blocked', function(done) {
```
tchx84: You can simplify this test if you provide a override file for the actual case you're aiming for. Otherwise thi becomes redundant and hard to read.

tests/src/testModels.js
```diff
@@ -1493,4 +1499,166 @@ describe('Model', function() {
...
+        expect(permissionsDefault.usb).toEqual('vnd:0123');
+
+        permissionsDefault.set_property('usb', 'vnd:0123;vnd:0456');
+
+        GLib.timeout_add(GLib.PRIORITY_HIGH, delay + 1, () => {
+            expect(has(_usbOverride, _usbGroup, _usbKey, 'vnd:0456')).toBe(true);
+
+            permissionsDefault.set_property('usb', 'vnd:0123');
+
+            GLib.timeout_add(GLib.PRIORITY_HIGH, delay + 1, () => {
```
tchx84: Same here, see if you can just re-use one of the existing overrides file you already added instead of doing this callback nesting.

tests/src/testModels.js
```diff
@@ -1493,4 +1499,166 @@ describe('Model', function() {
...
+        expect(permissionsDefault.usb_hidden).toEqual('vnd:0408+dev:5348');
+
+        permissionsDefault.set_property('usb-hidden', 'vnd:0408+dev:5348;vnd:0456');
+
+        GLib.timeout_add(GLib.PRIORITY_HIGH, delay + 1, () => {
+            expect(has(_usbOverride, _usbGroup, _usbHiddenKey, 'vnd:0456')).toBe(true);
+
+            permissionsDefault.set_property('usb-hidden', 'vnd:0408+dev:5348');
+
+            GLib.timeout_add(GLib.PRIORITY_HIGH, delay + 1, () => {
```
tchx84: Same as other suggestions.

src/models/usb.js
```diff
@@ -0,0 +1,220 @@
...
+        proxy.set_property('usb-status', this.constructor.serialize(usbStatuses));
+
+        const hiddenStatuses = this.constructor.deserialize(proxy.usb_hidden)
+            .filter(d => d.length !== 0)
+            .map(d => this._getHiddenStatusForDevice(d));
+        proxy.set_property('usb-hidden-status', this.constructor.serialize(hiddenStatuses));
+    }
+
+    updateProxyProperty(proxy) {
+        const allHidden = new Set([
```
tchx84: What's the difference between allHidden and blocked?

src/models/usb.js
```diff
@@ -0,0 +1,220 @@
...
+            status = FlatsealOverrideStatus.GLOBAL;
+        if (this._hiddenOverrides.has(device))
+            status = FlatsealOverrideStatus.USER;
+        return status;
+    }
+
+    updateFromProxyProperty(property, value) {
+        const devices = new Set(this.constructor.deserialize(value)
+            .filter(d => d.length !== 0));
+        const serialized = this.constructor.serialize([...devices]);
```
tchx84: I am not entirely sure I understand what this is workaround is solving, do you mind explaining with an example?

tchx84 (changes requested):
We're getting closer but, from now on, please only add changes that strictly address my comments. If you add other unrelated changes in between, we simply can't converge on a solution we're both satisfied with.

src/models/usb.js
Malika7188: Let me walk through exactly what happens.
**Setup:** `com.test.Usb` has `vnd:0123` as an originally allowed device, and `vnd:0408+dev:5348` as an originally blocked device.
**The user does one thing**: removes `vnd:0123` from the "Devices" (allowed) list. They never touch "Blocked devices" at all.

**The problem:** `_updateModels()` in `permissions.js` re-reads every property of every model on every update cycle, not just the one the user changed. So even though only usb changed, `updateFromProxyProperty('usb-hidden', ...)` still gets called too, with whatever usb-hidden's value currently is (`vnd:0408+dev:5348`, unchanged, since the user never touched it)

**Without this check:** here's what happens in one cycle:
1. **usb branch runs:** sees `vnd:0123` is missing from allowed then correctly adds it to `_hiddenOverrides `(blocks it).
2. **usb-hidden branch runs right after, same cycle:** it has no idea step 1 just happened. It just takes the current usb-hidden value and rebuilds `_hiddenOverrides` from scratch using that value. Since that value never included `vnd:0123`, the rebuild throws away the block from step 1.

`End result:` the override file ends up empty, gets deleted, and `vnd:0123 `quietly goes back to allowed, even though the user just removed it. That was the original bug. Both steps happen purely in memory, before anything is ever written to disk, the block never gets a chance to be saved, it's undone before the save even runs.

**What this check fixes:** before rebuilding, the usb-hidden branch first asks: has this value actually changed since I last saw it? If not, it skips the rebuild and leaves whatever usb just did alone. It only rebuilds when` usb-hidden` has genuinely changed, meaning the user actually edited the Blocked-devices list themselves.

**So in short:** this stops one property from being reprocessed with an old, unchanged value in a way that would undo what the other property just correctly did, in that same cycle.

tchx84: Thanks, that example shows this is just a minor logic limitation, and is not an unavoidable design flaw. So it can still be improved.

Consider the following facts:

1. Each time the user modifies permissions, that delayed callback will always run and process the current state of the UI (toggles, entries, etc).
2. The callback will always process all models.
3. The callback will always process both usb and usb-hidden properties.
4. The callback will always process usb first.
5. The callback will always process usb-hidden second.
6. The `updateFromProxyProperty` will only receive one property at the time in that order, and it doesn't need to process both properties at the same time.

Given these facts, let's reevaluate this statement again:

> usb-hidden branch runs right after, same cycle: it has no idea step 1 just happened. It just takes the current usb-hidden value and rebuilds _hiddenOverrides from scratch using that value. Since that value never included vnd:0123, the rebuild throws away the block from step 1.

This statement exposes the limitation; not knowing that step 1 just happened is just a self-imposed limitation, and throwing away what happened in step 1 is just a flaw in the logic.

Alternatively, how about doing this:

1. When usb is processed (step 1), it stores the hidden devices in a "pendingHiddenDevices" internal property.
2. When usb-hidden is processed (step 2), it adds whatever is in the "pendingHiddenDevices" into "_hiddenOverrides", and then resets "pendingHiddenDevices" it for the next cycle.

This way, step 2 is aware of what happened in step 1, and you simplify the overall logic. After all, this is basically just a 2 step model, which should work robustly given the facts that I outlined above.

tests/src/testModels.js
```diff
@@ -1583,25 +1587,21 @@ describe('Model', function() {
 
     it('restores auto-blocked original USB device to allowed when removed from blocked', function(done) {
         GLib.setenv('FLATPAK_USER_DIR', _tmp, true);
-        permissionsDefault.appId = _usbAppId;
 
-        expect(permissionsDefault.usb).toEqual('vnd:0123');
+        const keyFile = new GLib.KeyFile();
+        keyFile.set_value(_usbGroup, _usbHiddenKey, 'vnd:0123');
+        keyFile.save_to_file(_usbOverride);
```
tchx84: Closer, but what I meant is to provide these overrides as "fixtures", not to create these files in runtime.  

Basically, you can add a whole new directory structure under "tests/content" (e.g., `tests/content/usb/caseXYZ/`) for these overrides, tailored to each case, and activate the right override as usual using `GLib.setenv('FLATPAK_USER_DIR', _pathToTailoredOverrideForThisPaticularTest, true);`

With this you can also validate more clearly the original assumptions of each case, as I suggested.

tchx84 (changes requested):

tests/src/testModels.js
Malika7188: Thanks for the suggestion. I reworked the implementation following the two-step approach, and I think it's much simpler than my previous solution.

While testing it, I noticed a device was being re-added to the pending blocked list on every cycle, which prevented it from ever being unblocked. I added a check so it's only added once, and added a test to confirm it stays unblocked afterward.

I also noticed the same behavior needed to work whether a device was already blocked when the app opened or became blocked while it was running. I updated the implementation so both cases are handled correctly

I verified the behavior manually and ran the full test suite, which passes.

Malika7188: > Closer, but what I meant is to provide these overrides as "fixtures", not to create these files in runtime.
> 
> Basically, you can add a whole new directory structure under "tests/content" (e.g., `tests/content/usb/caseXYZ/`) for these overrides, tailored to each case, and activate the right override as usual using `GLib.setenv('FLATPAK_USER_DIR', _pathToTailoredOverrideForThisPaticularTest, true);`
> 
> With this you can also validate more clearly the original assumptions of each case, as I suggested.

I've refactored the USB override tests to use static fixture files as suggested, and all the tests pass.

While testing, I noticed that because `FLATPAK_USER_DIR` is used for both reading and writing overrides, running the test suite clears the data in those fixture files. The tests still pass, but after a local run, I had to manually add the fixture files back before running the suite again. I just wanted to flag this in case you'd prefer a different approach.

tests/src/testModels.js
```diff
@@ -1653,4 +1644,25 @@ describe('Model', function() {
         update();
     });
 
+    it('restore original USB device to allowed when auto-blocked live and later removed from blocked', function(done) {
```
tchx84: Can you clarify why this test was added and written with these nested callbacks?

src/models/usb.js
```diff
@@ -0,0 +1,210 @@
...
+        if (property === 'usb') {
+           
+
+            this._overrides = new Set([...devices]
+                .filter(d => !this._originals.has(d))
+                .filter(d => !this._globals.has(d)));
+
+                [...this._originals].forEach(d => {
+                if (!devices.has(d)) {
+                    if (!this._blockedOriginals.has(d))
```
tchx84: I am not sure I understand what's the role of this `blockedOriginals`. What's the difference between `hiddenOriginals` and `blockedOriginals`?

tchx84 (commented):
We're getting closer, but I need some clarifications with these usb<->usb-hidden interactions.

@Malika7188 can you summarize again the intended interactions between these two properties?

Malika7188:
> We're getting closer, but I need some clarifications with these usb<->usb-hidden interactions.
> 
> @Malika7188 can you summarize again the intended interactions between these two properties?

The Flatpak USB portal has no negation syntax (no `! `prefix like other permissions), I confirmed this against` flatpak-build-finish.xml`. So if an app's manifest originally grants `usb=vnd:0123` (enumerable), there's no way to remove that access by omission; the only way to actually revoke it is to add `vnd:0123` to hidden-devices instead, since hidden always takes precedence over enumerable.

That means the two lists aren't fully independent.For original devices: removing an original device from "Devices" (usb) has to auto-add it to "Blocked devices" (usb-hidden) to actually take effect. And the reverse: removing that same device later from "Blocked devices" should restore it to allowed, since it's still an original enumerable device that's simply no longer hidden.

For a device the user added themselves (not an original), whether in Devices or Blocked devices, removing it just deletes it directly. There's no automatic back-and-forth between the two lists, since it was never backed by an original permission that needs preserving.

So the auto-block/auto-restore behavior only applies to originals, user-added overrides on either list are just added and removed directly.

tests/src/testModels.js
Malika7188: > Can you clarify why this test was added and written with these nested callbacks?

This test checks the entire process: block a device, then later unblock it, and confirm it correctly returns to allowed. The other three tests start from a device that's already blocked; they only check the "unblock" half. This one instead does both steps, one after another: first remove `vnd:0123` from "Devices" (which should get it auto-blocked), then, once that's done, remove it from "Blocked devices" (which should restore it).

The reason I added a wait inside another wait is that Flatseal doesn't act on a change immediately. It waits about half a second after each change before actually applying it. So the test has to wait for the first change to finish being applied before making the second change. If it didn't wait, the second change would happen too soon and probably cancel or merge with the first one, and we'd end up testing something different from what actually happens when a user clicks Block, waits, then clicks Unblock.

src/models/usb.js
Malika7188: > I am not sure I understand what's the role of this `blockedOriginals`. What's the difference between `hiddenOriginals` and `blockedOriginals`?

`_hiddenOriginals ` is data loaded straight from the app's manifest, the devices the app itself already declares as hidden in hidden-devices. It's loaded once, same role as _originals plays for the enumerable side.

`_blockedOriginals` is different, it's not loaded from anywhere, it's recalculated every time `updateProxyProperty` runs. it keeps track of the devices that the app originally allowed, which are currently blocked. It's not permission data itself, it's bookkeeping.

**The reason it exists:**
The USB processing step checks whether originally-allowed devices are still in the allowed list. Without _blockedOriginals, any missing device would be treated as newly blocked on every cycle, causing already-blocked devices, including ones the user just unblocked to be re-added and the unblock action to be silently undone.

`_blockedOriginals` is what lets the code distinguish between "already accounted for as blocked" and "just went missing this cycle and needs blocking", so a device only gets flagged in the cycle where it actually transitions, not on every cycle afterward. That's the bug the test above is checking doesn't happen.

Malika7188:
I investigated the Environment variable regression further and found what I believe is the actual root cause.

Initially, I thought the problem was the regex. The original implementation only accepted complete `KEY=VALUE` entries, so I assumed an incomplete entry like `G` was being rejected by `VAR_REGEXP` and disappearing.  I changed the model to preserve incomplete entries instead of discarding them.

However, after making that change, the regression still happened, and I investigated further.

Suppose an application already has:

`TEST=yes`

If I add a new row and start typing a new variable name, for example `G`, the new row immediately disappears and the `G` gets appended to the previous variable, so it becomes:

`TEST=yes;G`

From what I found, this happens before the regex is even involved.

The UI sends:

`TEST=yes;G`

to the model, and `deserialize()` tries to split it back into individual variables. It only recognizes a new variable when it sees `;NAME=`. Since `G` doesn't have an `=` yet, it isn't recognized as the start of a new variable. Instead, it's treated as part of the value of `TEST`.

By the time the regex runs, it only sees one complete entry:

`TEST=yes;G`

which is valid because environment variable values are allowed to contain semicolons. So the regex never gets a chance to classify `G` as an incomplete variable because it has already been merged into the previous one.

When `_updateProperties()` runs, it simply redraws what the model interpreted, so the UI shows a single row (`TEST=yes;G`) and the new row disappears.

So my initial diagnosis about the regex was only part of the picture. Fixing the regex gate solved the problem of incomplete entries being discarded, but the deeper issue is that the model never recognizes the unfinished variable as a separate entry in the first place because of how `deserialize()` reconstructs the variables from the serialized string.

tchx84:
> I investigated the Environment variable regression further and found what I believe is the actual root cause.
> 
> Initially, I thought the problem was the regex. The original implementation only accepted complete `KEY=VALUE` entries, so I assumed an incomplete entry like `G` was being rejected by `VAR_REGEXP` and disappearing. I changed the model to preserve incomplete entries instead of discarding them.
> 
> However, after making that change, the regression still happened, and I investigated further.
> 
> Suppose an application already has:
> 
> `TEST=yes`
> 
> If I add a new row and start typing a new variable name, for example `G`, the new row immediately disappears and the `G` gets appended to the previous variable, so it becomes:
> 
> `TEST=yes;G`
> 
> From what I found, this happens before the regex is even involved.
> 
> The UI sends:
> 
> `TEST=yes;G`
> 
> to the model, and `deserialize()` tries to split it back into individual variables. It only recognizes a new variable when it sees `;NAME=`. Since `G` doesn't have an `=` yet, it isn't recognized as the start of a new variable. Instead, it's treated as part of the value of `TEST`.
> 
> By the time the regex runs, it only sees one complete entry:
> 
> `TEST=yes;G`
> 
> which is valid because environment variable values are allowed to contain semicolons. So the regex never gets a chance to classify `G` as an incomplete variable because it has already been merged into the previous one.
> 
> When `_updateProperties()` runs, it simply redraws what the model interpreted, so the UI shows a single row (`TEST=yes;G`) and the new row disappears.
> 
> So my initial diagnosis about the regex was only part of the picture. Fixing the regex gate solved the problem of incomplete entries being discarded, but the deeper issue is that the model never recognizes the unfinished variable as a separate entry in the first place because of how `deserialize()` reconstructs the variables from the serialized string.

This regression affects all the "entry" permissions. The root cause *is* that *_updateProperties* you added and should be removed, but it's important that *you* understand why you added it in the first place to truly understand this whole circular-behavior issue.

src/models/usb.js
```diff
@@ -0,0 +1,188 @@
...
+    updateFromProxyProperty(property, value) {
+        const devices = new Set(this.constructor.deserialize(value)
+            .filter(d => d.length !== 0));
+
+        if (property === 'usb') {
+            this._overrides = new Set([...devices]
+                .filter(d => !this._originals.has(d))
+                .filter(d => !this._globals.has(d)));
+        } else if (property === 'usb-hidden') {
+            const hiddenKnown = new Set([...this._hiddenOriginals, ...this._hiddenGlobals]);
```
tchx84: This  'usb-hidden' block would be easier to read if it did the same thing as the 'usb' block. e.g., we probably don't really need this 'hiddenKnown' variable right?

src/widgets/usbRow.js
```diff
@@ -0,0 +1,118 @@
...
+        this._entry.set_text(text);
+    }
+
+    get status() {
+        return this._statusIcon.status;
+    }
+
+    set status(status) {
+        this._statusIcon.status = status;
+        this._button.sensitive = status === FlatsealOverrideStatus.USER;
```
tchx84: Better to add a brief comment here about what and why we're doing here.

tests/src/testModels.js
```diff
@@ -67,6 +68,9 @@ const _global = GLib.build_filenamev(['..', 'tests', 'content', 'global', 'flatp
 const _globalNegated = GLib.build_filenamev(['..', 'tests', 'content', 'globalNegated', 'flatpak']);
 const _globalResetMode = GLib.build_filenamev(['..', 'tests', 'content', 'globalResetMode', 'flatpak']);
 const _statuses = GLib.build_filenamev(['..', 'tests', 'content', 'statuses', 'flatpak']);
+const _usbBlockedOriginal = GLib.build_filenamev(['..', 'tests', 'content', 'usb', 'blockedOriginal', 'flatpak']);
+const _usbUserAllowed = GLib.build_filenamev(['..', 'tests', 'content', 'usb', 'userAllowed', 'flatpak']);
+const _usbUserBlocked = GLib.build_filenamev(['..', 'tests', 'content', 'usb', 'userBlocked', 'flatpak']);
```
tchx84: These two `_usb*` variables don't need to exist anymore right? You removed these files. Please double check if there other left overs to be removed.

tests/src/testModels.js
```diff
@@ -1493,4 +1505,136 @@ describe('Model', function() {
...
+            expect(hasInTotal(_usbOverride)).toEqual(2);
+            done();
+            return GLib.SOURCE_REMOVE;
+        });
+
+        update();
+    });
+
+
+    it('restores auto-blocked original USB device to allowed when removed from blocked', function(done) {
```
tchx84: Didn't we remove all "auto" behavior? Why this is still here?

tests/src/testModels.js
```diff
@@ -1493,4 +1505,136 @@ describe('Model', function() {
...
+        GLib.timeout_add(GLib.PRIORITY_HIGH, delay + 1, () => {
+            expect(has(_usbOverride, _usbGroup, _usbHiddenKey, 'vnd:0123')).toBe(true);
+            permissionsDefault.appId = _usbAppId;
+            expect(permissionsDefault.usb).not.toContain('vnd:0123');
+            done();
+            return GLib.SOURCE_REMOVE;
+        });
+        update();
+    });
+    it('removes user-added USB device from allowed without blocking it', function(done) {
```
tchx84: I see a lot of tests that don't need to exist anymore, please only keep necessary tests.

NOTES.md
```diff
@@ -0,0 +1,17 @@
+different shell for mobile apps
```
tchx84: No need for this README.

tchx84 (changes requested):
A few things to improve, but a lot of leftovers to be removed. Please check carefully what is not needed anymore.

Additionally, regarding formatting perhaps you can run ESLint on the new files you added? Flatseal includes a `.eslintrc.yml` file, there might still be a way to reuse it with newer ESLint versions.

src/models/usb.js
```diff
@@ -0,0 +1,188 @@
...
+        const blocked = new Set([
+            ...this._hiddenOriginals,
+            ...this._hiddenGlobals,
+            ...this._hiddenOverrides,
+        ]);
+
+        const allowed = new Set([
+            ...[...this._originals].filter(d => !blocked.has(d)),
+            ...[...this._globals].filter(d => !blocked.has(d)),
+            ...[...this._overrides].filter(d => !blocked.has(d)),
```
tchx84: Now that the two properties are independent, why is this being filtered by "blocked"?

src/models/usb.js
```diff
@@ -0,0 +1,188 @@
...
+    _init() {
+        super._init({});
+        this.reset();
+    }
+
+    getPermissions() {
+        return {
+            usb: {
+                supported: this._info.supports('1.15.11'),
+                description: _('Devices'),
```
tchx84: Maybe be more explicit, like "Allowed Devices"?

tchx84 (changes requested):

src/models/usb.js
Malika7188: Thanks for catching that. Fixed.

tchx84 (changes requested):
hey @Malika7188 , this is almost ready but I found a few more details:

* Your commit message doesn't include your full name. Is that intentional?
* Although you can't remove non-user-changed rows with the button, you can still edit the text of these rows. Please fix this and test the feature a bit more (manually).
* The documentation entry for "hidden" does not mention that it takes precedence over the allow list. Which is critical information, given how this implementation works.

Malika7188:
Hi @tchx84, 
I've made the changes: the entire USB row is now disabled for non-user entries instead of just the remove button, which also fixes the text being editable. I've tested this manually. 
I've also updated both documentation files (`DOCUMENTATION.md` and `help/C/index.html`) to note that a blocked device takes precedence over the allow list when it exists in both, and fixed the commit author to show my full name.

PR #885, opened 2026-08-10
models: Fix crash from invalid release timestamps

Malika7188:
An application's metainfo could contain a release timestamp too large
or too small to represent as a valid date. Formatting such a date could
throw an uncaught error during startup and prevent Flatseal from launching.

Check the date before formatting it and fall back to the existing default
when the timestamp is invalid.


strobelight (approved):
This worked for me. Flatseal launched and showed "Unknown" for "Last Updated".

I built a flatpak with metainfo release date of "1969-12-31".

Malika7188:
This PR and the related unguarded parsing issues are now addressed in #888

PR #886, opened 2026-08-15
Add conditional permission support

Malika7188:
Previously, conditional permission entries (if:...) were discarded
entirely while parsing, to avoid treating them as unsupported and
corrupting the overrides file.

This now loads the underlying permission through the normal path,
the same as any other option, and separately records the raw
condition. A new status icon marks a permission row as conditional,
with a tooltip stating the condition it depends on.

**Limitations**
Flatpak drops a conditional permission if a bare grant for the same 
permission is applied afterward, so the switch showing "on" doesn’t 
always mean the permission is actually granted at runtime. 
This is Flatpak’s behaviour and not something Flatseal controls.

 `saveToKeyFile` right now only writes from the `overrides` Set (the on/off state)
 and doesn’t look at the conditionals map. So saving can drop an existing 
`conditional` entry from the override file. Write-back is not implemented yet

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -258,4 +258,4 @@ globals:
   beforeAll: false
   afterAll: false
 parserOptions:
-  ecmaVersion: 2018
+  ecmaVersion: 2020
--- a/src/models/permissions.js
+++ b/src/models/permissions.js
@@ -204,27 +204,6 @@ var FlatpakPermissionsModel = GObject.registerClass({
                     .replace(/;+$/, '')
                     .split(';');
 
-                /* Group entries by the plain option name they belong to, so a
-                 * conditional's full original text can be preserved
-                 * on write-back. */
-                const optionGroups = new Map();
-
-                values.forEach(option => {
-                    let plainOption = option.replace('!', '');
-
-                    if (option.startsWith(CONDITIONAL_PREFIX)) {
-                        const parts = option.slice(CONDITIONAL_PREFIX.length).split(':');
-
-                        if (parts.length >= 2)
-                            [plainOption] = parts;
-                    }
-
-                    if (!optionGroups.has(plainOption))
-                        optionGroups.set(plainOption, []);
-
-                    optionGroups.get(plainOption).push(option);
-                });
-
                 values.forEach(option => {
                     let isConditional = false;
                     let bareOption = option;
@@ -232,6 +211,10 @@ var FlatpakPermissionsModel = GObject.registerClass({
                     if (option.startsWith(CONDITIONAL_PREFIX)) {
                         const parts = option.slice(CONDITIONAL_PREFIX.length).split(':');
 
+                        /* A valid conditional has both an option and a
+                         * condition after "if:". Ignore incomplete
+                         * entries instead of treating them as
+                         * conditionals. */
                         if (parts.length >= 2) {
                             isConditional = true;
                             [bareOption] = parts;
@@ -243,19 +226,22 @@ var FlatpakPermissionsModel = GObject.registerClass({
                     if (model === null)
                         model = this.constructor._find(`${group}_${key}`);
 
-                    if (model === null && overrides && !global && !isConditional)
+                    if (model === null && overrides && !global)
                         model = MODELS.unsupported;
 
-                    /* Ignore conditional entries that don't match a
-                     * supported model. */
+                    /* Only the four models in CONDITIONAL_MODELS are
+                     * recognized to support conditionals. A conditional
+                     * entry for anything else is skipped entirely here:
+                     * not loaded into any model, not shown in the UI,
+                     * and never tracked, so nothing is ever written
+                     * back for it either. This avoids guessing at
+                     * syntax that is not supported and risking
+                     * corruption of the override file. */
                     if (isConditional && CONDITIONAL_MODELS.includes(model)) {
-                        const plainOption = bareOption.replace('!', '');
-                        const fullGroup = optionGroups.get(plainOption).join(';');
-
                         model?.loadFromKeyFile(group, key, bareOption, overrides, global);
-                        model?.markConditional(plainOption, option, fullGroup, overrides, global);
+                        model?.markConditional(bareOption, option);
                     } else if (!isConditional) {
-                        model?.loadFromKeyFile(group, key, option, overrides, global);
+                        model?.loadFromKeyFile(group, key, bareOption, overrides, global);
                     }
                 });
             });
--- a/src/models/permissions.js
+++ b/src/models/permissions.js
@@ -53,7 +53,12 @@ const MODELS = {
 };
 
 /* Models that support conditional permissions */
-const CONDITIONAL_MODELS = [MODELS.shared, MODELS.sockets, MODELS.devices, MODELS.features];
+const CONDITIONAL_MODELS = [
+    MODELS.shared,
+    MODELS.sockets,
+    MODELS.devices,
+    MODELS.features,
+];
 
 function generate_index() {
     const index = {};
@@ -216,17 +221,14 @@ var FlatpakPermissionsModel = GObject.registerClass({
                     if (model === null && overrides && !global)
                         model = MODELS.unsupported;
 
-                    if (model === null)
-                        return;
-
                     /* Preserves the original conditional string for models
                      * that don't support conditionals, so the condition
                      * is not lost.*/
-                    if (isConditional && model !== MODELS.unsupported) {
-                        model.loadFromKeyFile(group, key, bareOption, overrides, global);
-                        model.markConditional(bareOption, option);
+                    if (isConditional && CONDITIONAL_MODELS.includes(model)) {
+                        model?.loadFromKeyFile(group, key, bareOption, overrides, global);
+                        model?.markConditional(bareOption, option);
                     } else {
-                        model.loadFromKeyFile(group, key, option, overrides, global);
+                        model?.loadFromKeyFile(group, key, option, overrides, global);
                     }
                 });
             });
--- a/src/models/permissions.js
+++ b/src/models/permissions.js
@@ -204,6 +204,27 @@ var FlatpakPermissionsModel = GObject.registerClass({
                     .replace(/;+$/, '')
                     .split(';');
 
+                /* Group entries by the plain option name they belong to, so a
+                 * conditional's full original text can be preserved
+                 * on write-back. */
+                const optionGroups = new Map();
+
+                values.forEach(option => {
+                    let plainOption = option.replace('!', '');
+
+                    if (option.startsWith(CONDITIONAL_PREFIX)) {
+                        const parts = option.slice(CONDITIONAL_PREFIX.length).split(':');
+
+                        if (parts.length >= 2)
+                            [plainOption] = parts;
+                    }
+
+                    if (!optionGroups.has(plainOption))
+                        optionGroups.set(plainOption, []);
+
+                    optionGroups.get(plainOption).push(option);
+                });
+
                 values.forEach(option => {
                     let isConditional = false;
                     let bareOption = option;
@@ -228,8 +249,11 @@ var FlatpakPermissionsModel = GObject.registerClass({
                     /* Ignore conditional entries that don't match a
                      * supported model. */
                     if (isConditional && CONDITIONAL_MODELS.includes(model)) {
+                        const plainOption = bareOption.replace('!', '');
+                        const fullGroup = optionGroups.get(plainOption).join(';');
+
                         model?.loadFromKeyFile(group, key, bareOption, overrides, global);
-                        model?.markConditional(bareOption, option);
+                        model?.markConditional(plainOption, option, fullGroup, overrides, global);
                     } else if (!isConditional) {
                         model?.loadFromKeyFile(group, key, option, overrides, global);
                     }
--- a/src/models/permissions.js
+++ b/src/models/permissions.js
@@ -52,6 +52,9 @@ const MODELS = {
     unsupported: new FlatpakUnsupportedModel(),
 };
 
+/* Models that support conditional permissions */
+const CONDITIONAL_MODELS = [MODELS.shared, MODELS.sockets, MODELS.devices, MODELS.features];
+
 function generate_index() {
     const index = {};
 
@@ -197,24 +200,15 @@ var FlatpakPermissionsModel = GObject.registerClass({
                     .split(';');
 
                 values.forEach(option => {
+                    let isConditional = false;
+                    let bareOption = option;
+
                     if (option.startsWith(CONDITIONAL_PREFIX)) {
-                        /* Record the conditional request as a real
-                         * permission, without letting it fall into the
-                         * unsupported bucket. */
-                        const [conditionalOption] = option.slice(CONDITIONAL_PREFIX.length).split(':');
-                        let conditionalModel = this.constructor._find(`${group}_${key}_${conditionalOption}`);
-
-                        if (conditionalModel === null)
-                            conditionalModel = this.constructor._find(`${group}_${key}`);
-
-                        if (conditionalModel !== null) {
-                            conditionalModel.loadFromKeyFile(group, key, conditionalOption, overrides, global);
-                            conditionalModel.markConditional(conditionalOption, option);
-                        }
-                        return;
+                        isConditional = true;
+                        [bareOption] = option.slice(CONDITIONAL_PREFIX.length).split(':');
                     }
 
-                    model = this.constructor._find(`${group}_${key}_${option.replace('!', '')}`);
+                    model = this.constructor._find(`${group}_${key}_${bareOption.replace('!', '')}`);
 
                     if (model === null)
                         model = this.constructor._find(`${group}_${key}`);
@@ -222,8 +216,18 @@ var FlatpakPermissionsModel = GObject.registerClass({
                     if (model === null && overrides && !global)
                         model = MODELS.unsupported;
 
-                    if (model !== null)
+                    if (model === null)
+                        return;
+
+                    /* Preserves the original conditional string for models
+                     * that don't support conditionals, so the condition
+                     * is not lost.*/
+                    if (isConditional && model !== MODELS.unsupported) {
+                        model.loadFromKeyFile(group, key, bareOption, overrides, global);
+                        model.markConditional(bareOption, option);
+                    } else {
                         model.loadFromKeyFile(group, key, option, overrides, global);
+                    }
                 });
             });
         });
@@ -287,8 +291,7 @@ var FlatpakPermissionsModel = GObject.registerClass({
         GObject.signal_handler_block(this, this._notifyHandlerId);
 
         Object.values(MODELS).forEach(model => model.updateStatusProperty(this));
-        [MODELS.shared, MODELS.sockets, MODELS.devices, MODELS.features]
-            .forEach(model => model.updateConditionalProperty(this));
+        CONDITIONAL_MODELS.forEach(model => model.updateConditionalProperty(this));
 
         GObject.signal_handler_unblock(this, this._notifyHandlerId);
     }
--- a/src/models/permissions.js
+++ b/src/models/permissions.js
@@ -209,8 +209,12 @@ var FlatpakPermissionsModel = GObject.registerClass({
                     let bareOption = option;
 
                     if (option.startsWith(CONDITIONAL_PREFIX)) {
-                        isConditional = true;
-                        [bareOption] = option.slice(CONDITIONAL_PREFIX.length).split(':');
+                        const parts = option.slice(CONDITIONAL_PREFIX.length).split(':');
+
+                        if (parts.length >= 2) {
+                            isConditional = true;
+                            [bareOption] = parts;
+                        }
                     }
 
                     model = this.constructor._find(`${group}_${key}_${bareOption.replace('!', '')}`);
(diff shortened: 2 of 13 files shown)


src/models/permissions.js
```diff
@@ -192,10 +197,22 @@ var FlatpakPermissionsModel = GObject.registerClass({
...
+                         * unsupported bucket. */
+                        const [conditionalOption] = option.slice(CONDITIONAL_PREFIX.length).split(':');
+                        let conditionalModel = this.constructor._find(`${group}_${key}_${conditionalOption}`);
+
+                        if (conditionalModel === null)
+                            conditionalModel = this.constructor._find(`${group}_${key}`);
+
+                        if (conditionalModel !== null) {
+                            conditionalModel.loadFromKeyFile(group, key, conditionalOption, overrides, global);
+                            conditionalModel.markConditional(conditionalOption, option);
```
tchx84: Is this whole parallel path really needed? Can't we just "extract" the permission out of the condition, treat it as usual and then only markConditional? So the special treatment is minimal for these.

src/models/permissions.js
```diff
@@ -270,6 +287,8 @@ var FlatpakPermissionsModel = GObject.registerClass({
         GObject.signal_handler_block(this, this._notifyHandlerId);
 
         Object.values(MODELS).forEach(model => model.updateStatusProperty(this));
+        [MODELS.shared, MODELS.sockets, MODELS.devices, MODELS.features]
```
tchx84: Add a CONDITIONAL_MODELS list so it's easy to keep track of that list? Plus, here it will end up looking like the preceding line.

tchx84 (changes requested):
Besides the other comments, in order to design this in a future-proof / defensive way, imagine if tomorrow some other permission adds support for conditional but Flatseal doesn't support it, what Flatseal's behavior should be?

Malika7188:
> Besides the other comments, in order to design this in a future-proof / defensive way, imagine if tomorrow some other permission adds support for conditional but Flatseal doesn't support it, what Flatseal's behavior should be?

I’ve fixed this. If a future permission adds conditional support that Flatseal doesn’t recognize yet, it now falls into the same unsupported bucket as a normal unrecognized permission and is preserved as it is, including the `if: prefix.` This way, it can be written back unchanged when saving instead of losing the condition.

src/models/permissions.js
```diff
@@ -192,21 +200,34 @@ var FlatpakPermissionsModel = GObject.registerClass({
...
 
                 values.forEach(option => {
-                    /* Flatseal does not support conditionals, but skips them
-                     * to avoid corrupting the overrides file. */
-                    if (option.startsWith(CONDITIONAL_PREFIX))
-                        return;
+                    let isConditional = false;
+                    let bareOption = option;
+
+                    if (option.startsWith(CONDITIONAL_PREFIX)) {
```
tchx84, line 211: A couple things to double check:

1. The conditional permission syntax, is there another syntax that could break here? Update to latest flatpak v1.18 and check `man flatpak-metadata`.
2. Why not also checking if the model is in the supported list here?
3. Would't be "safer" to just ignore the raw conditional options that aren't explicitly supported? (might be related to alternative conditional syntax).

src/models/permissions.js
```diff
@@ -52,6 +52,9 @@ const MODELS = {
     unsupported: new FlatpakUnsupportedModel(),
 };
 
+/* Models that support conditional permissions */
+const CONDITIONAL_MODELS = [MODELS.shared, MODELS.sockets, MODELS.devices, MODELS.features];
```
tchx84: Nitpick; break each list item in a separate line (with a trailing comma).

src/models/permissions.js
```diff
@@ -192,21 +200,34 @@ var FlatpakPermissionsModel = GObject.registerClass({
...
+                    model = this.constructor._find(`${group}_${key}_${bareOption.replace('!', '')}`);
 
                     if (model === null)
                         model = this.constructor._find(`${group}_${key}`);
 
                     if (model === null && overrides && !global)
                         model = MODELS.unsupported;
 
-                    if (model !== null)
+                    if (model === null)
```
tchx84: Depending on the previous double checks, this whole part of the change can be reduced to just marking the permission as optional.

tests/src/testModels.js
```diff
@@ -597,6 +597,7 @@ describe('Model', function() {
             expect(has(_unsupportedOverride, 'Context', 'unsupported', 'always')).toBe(true);
             expect(has(_unsupportedOverride, 'Context', 'unsupported', 'undefined')).toBe(false);
             expect(has(_unsupportedOverride, 'Context', 'unsupported', 'null')).toBe(false);
+            expect(has(_unsupportedOverride, 'Context', 'unsupported', 'if:teleport:true')).toBe(true);
```
tchx84: Can you double check what happens if someone uses a different conditional syntax ? Keep in mind that changing the ordering/location of the conditional component, might have unexpected effects.

tchx84 (changes requested):

src/models/permissions.js
Malika7188: Optional chaining works fine at runtime. The issue is that `.eslintrc.yml` had `ecmaVersion` set to 2018, so ESLint couldn't parse `?.` and I was getting:

`228:31  error  Parsing error: Unexpected token .`

I bumped `ecmaVersion` to 2020 and verified the project linting. It doesn't introduce any new lint issues.

Should I keep the `ecmaVersion` set to 2020 to support the optional chaining? Or you'd prefer a different approach?

tests/src/testModels.js
Malika7188: I tested this with a few different malformed cases, including `if:!has-wayland:x11` with the condition and permission swapped. None of them crash, and they’re preserved as unsupported entries even after saving.

I also tested missing conditions, extra colons, and a completely made-up permission name. All of them behaved the same way.

src/models/permissions.js
Malika7188: > 3\. Would't be "safer" to just ignore the raw conditional options that aren't explicitly supported? (might be related to alternative conditional syntax).

I've thought about this more. The key thing is that Flatseal doesn't touch the app's own metadata, only the override files. So this is really about something a user has put in the override file with `flatpak override` that Flatseal doesn't recognize yet.

The permission could already be valid and used by Flatpak. For example, if Flatpak adds conditional support for a permission before Flatseal adds a model for it. In that case, Flatseal simply doesn't know how to represent it yet.

If we ignore it, Flatseal would silently remove it the next time the overrides are saved. If we preserve it, it stays intact and can be handled normally once Flatseal adds support for that permission.

So I think preserving it is safer. We're keeping an override that Flatpak may already understand and use, even if Flatseal doesn't support it yet. 

What do you think? would preserving it be a better approach than ignoring it?

Malika7188: > 1. The conditional permission syntax, is there another syntax that could break here? Update to latest flatpak v1.18 and check `man flatpak-metadata`.

I also checked this against the latest Flatpak source. The checkout I used is `1.19.0` development, so it's newer than `1.18.` I checked the conditional permissions docs and the parser, and the syntax is still `if:PERMISSION:CONDITION`. I also checked the 1.18 and current changelogs, and there don't seem to be any changes to the conditional syntax since it was added in `1.17.`

I checked `man flatpak-metadata` too, but the man page available on my system is from Flatpak 1.16.6, so it doesn't include the conditional permissions section. The latest Flatpak source does include the updated documentation, so I used that to verify the syntax.

tchx84: > What do you think? would preserving it be a better approach than ignoring it?

The question is not whether we can preserve the "string", but whether we can preserve it in a way that's semantically equivalent (e.g., what happens if we mess up the ordering? like putting the string before something else and this changing the meaning).

Modifying the behavior indirectly (by mistake) is also a pretty bad outcome.

Malika7188: I tested this properly before answering, and it turned out to be two separate findings.

First, the ordering question you asked about. I ran several malformed/unsupported conditionals through a load-then-save cycle:

`BEFORE: shared=if:x11;if:!has-wayland:x11;if:x11:cond:extra:parts;if:teleport:true`
`AFTER: shared=if:teleport:true;if:x11:cond:extra:parts;if:!has-wayland:x11;if:x11;ipc`

The strings themselves are preserved, but their order changes because Flatseal stores them in a `Set`, which doesn't keep the original order. For these unsupported entries, that doesn't change their meaning because Flatpak evaluates each option separately, so moving unrelated entries around doesn't change what gets granted.

But then I tested the same load-then-save cycle with a real supported permission, using the documented backward-compatible pattern:

`BEFORE: sockets=!x11;x11;if:x11:!has-wayland;`
`AFTER: sockets=!x11`

Here the conditional isn't just reordered, it disappears completely. That's the same save limitation we already identified: `saveToKeyFile()` only writes the normal on/off state and doesn't write back the conditionals stored separately. So for a real permission, saving it currently drops the conditional.

Because of this, saving the file can quietly change what a permission actually does, without anyone touching it directly. So for unrecognized conditionals, I went with your suggestion and changed it to ignore them rather than keep something that may not be preserved correctly.

I also found another issue while testing malformed cases: `if:x11` (missing the condition) was being accepted because `x11` matched a real permission, even though the conditional itself is incomplete. Flatpak actually rejects this. I've added a check so incomplete conditionals aren't treated as valid permissions and instead go through the unrecognized path.

Does this match what you had in mind, or would you prefer it to be handled differently?

src/models/permissions.js
```diff
@@ -102,6 +110,11 @@ function generate() {
             const statusProperty = `${property}-status`;
             properties[statusProperty] = GObject.ParamSpec.string(
                 statusProperty, statusProperty, statusProperty, FLAGS, FlatsealOverrideStatus.ORIGINAL);
+
+            /* conditional requests */
+            const conditionalProperty = `${property}-conditional`;
+            properties[conditionalProperty] = GObject.ParamSpec.string(
+                conditionalProperty, conditionalProperty, conditionalProperty, FLAGS, '');
```
tchx84, line 117: Just to inform us, please check if there's a technical limit on how many property a GObject can export 🙏.

src/models/permissions.js
```diff
@@ -191,22 +204,59 @@ var FlatpakPermissionsModel = GObject.registerClass({
...
+
+                        if (parts.length >= 2)
+                            [plainOption] = parts;
+                    }
+
+                    if (!optionGroups.has(plainOption))
+                        optionGroups.set(plainOption, []);
+
+                    optionGroups.get(plainOption).push(option);
+                });
```
tchx84: This is not the right place to handle this edge case. Let's try to make only minimal changes here and move actual logic to the individual permissions models. And if it gets too complex, I rather just drop these for now.

src/models/shared.js
```diff
@@ -186,27 +206,99 @@ var FlatpakSharedModel = GObject.registerClass({
...
+
+        let value = this.constructor.getDefault();
+
+        if (permissions.has(option))
+            value = true;
+        if (permissions.has(`!${option}`))
+            value = false;
+
+        return value;
+    }
```
tchx84: This is a good attempt, but is adding a lot of complexity for a edge case. If we can't find a simple solution for the edge case, let's try to land the simpler implementation first.

tchx84 (changes requested):

src/models/permissions.js
Malika7188: I have checked the GLib and GObject headers. Property IDs use `guint`, and the property registration APIs also use `guint` for the property IDs and counts. Flatseal currently has 43 permissions, so with the conditional properties that's 129 properties on the model. I don't think the extra properties should be a concern

src/models/shared.js
Malika7188: I reverted the changes back to the previous simpler implementation, so the additional logic for preserving conditionals is no longer included.

tests/content/user/flatpak/overrides/com.test.Unsupported
```diff
@@ -1,3 +1,3 @@
 [Context]
 shared=unsupported
-unsupported=always
+unsupported=always;if:teleport:true
```
tchx84: I think there are more angles to this, is not only that `teleport` isn't supported, but it could also be `has-teleport`.

BTW, instead of random word, it make it easier to read/understand later if you use something more meaningful like `if:unsupported-permission:!has-unsupported-permission`.

.eslintrc.yml
```diff
@@ -258,4 +258,4 @@ globals:
   beforeAll: false
   afterAll: false
 parserOptions:
-  ecmaVersion: 2018
+  ecmaVersion: 2020
```
tchx84, line 261: Can you remind why is this needed still?

tests/src/testModels.js
```diff
@@ -1467,6 +1484,17 @@ describe('Model', function() {
...
 
+    it('marks conditional permissions', function() {
+        GLib.setenv('FLATPAK_USER_DIR', _user, true);
+        permissionsDefault.appId = _conditionalAppId;
+
+        expect(permissionsDefault.sockets_x11_conditional).toBe('if:x11:!has-wayland');
+        expect(permissionsDefault.devices_all_conditional).toBe('if:all:!has-input-device');
+
+        expect(permissionsDefault.sockets_wayland).toBe(true);
+        expect(permissionsDefault.sockets_wayland_conditional).toBe('if:wayland:true');
```
tchx84, line 1497: Please do run these tests.

src/models/permissions.js
```diff
@@ -192,21 +205,44 @@ var FlatpakPermissionsModel = GObject.registerClass({
...
+                     * and never tracked, so nothing is ever written
+                     * back for it either. This avoids guessing at
+                     * syntax that is not supported and risking
+                     * corruption of the override file. */
+                    if (isConditional && CONDITIONAL_MODELS.includes(model)) {
+                        model?.loadFromKeyFile(group, key, bareOption, overrides, global);
+                        model?.markConditional(bareOption, option);
+                    } else if (!isConditional) {
+                        model?.loadFromKeyFile(group, key, option, overrides, global);
+                    }
```
tchx84, line 245: It's not clear why there are two different `loadFromKeyFile` here, given `bareOption` defaults to `option`.

src/models/permissions.js
```diff
@@ -192,21 +205,44 @@ var FlatpakPermissionsModel = GObject.registerClass({
...
 
-                    model = this.constructor._find(`${group}_${key}_${option.replace('!', '')}`);
+                    model = this.constructor._find(`${group}_${key}_${bareOption.replace('!', '')}`);
 
                     if (model === null)
                         model = this.constructor._find(`${group}_${key}`);
 
-                    if (model === null && overrides && !global)
+                    if (model === null && overrides && !global && !isConditional)
```
tchx84: Similarly to my previous comment, why is this extra `!isConditional` check needed here?

Can't all the changes here be reduce to:

1) extract option from conditional and determine is a conditional.
2) mark option as conditional.

Instead of all this special logic?

tchx84 (changes requested):

.eslintrc.yml
Malika7188: I bumped `ecmaVersion` to 2020 because we use optional chaining in `permissions.js`. With `ecmaVersion` set to 2018, ESLint could not parse `?.` and threw a parsing error.

tests/src/testModels.js
Malika7188: I ran the tests, and `marks conditional permissions` passes. This tests whether Flatseal correctly remembers conditional permissions when loading an app's metadata or a per-app override, while still setting the actual permission value correctly so that the condition can be shown in the UI.

src/models/permissions.js
Malika7188: Fixed, thanks for catching that

src/models/permissions.js
Malika7188: I have removed `!isConditional` and verified the full test suite still passes. 

Also, these two steps are already happening in the code: the option is extracted from the conditional and the conditional is identified in `permissions.js` lines 208-222, then option is marked as conditional on line 242. Or did you have a different implementation in mind?

.eslintrc.yml
tchx84: Right, I am not against bumping this, but do we really need to use optional chaining in the first place?

I keep looking at the changes in `permissions.js`, and I would like you to analyze if we can change it with something that simply ends with:

```
if (model !== null) {
    model.loadFromKeyFile(group, key, bareOption, overrides, global);
    model.updateConditional(bareOption, option);
}
``` 

The idea goal is to reduce the amount of changes and branching we're adding there, and make it easier to understand.

Also, if we have to ignore a particular value, it would be better to do it explicitly at the beginning of that block and not silently as a consequence.

tchx84 (changes requested):

Malika7188:
Hi @tchx84 , I've made the changes and removed the optional chaining. The ignore unsupported conditionals case is now its own check at the top of the block. I also reverted `ecmaVersion` back to 2018 in `.eslintrc.yml`. Ran the checks and everything still passes.

Also, I misunderstood your earlier comment about "marking the permission as optional." I thought you meant using optional chaining (?.), that's why I used it. Sorry about that confusion.

tests/content/user/flatpak/overrides/com.test.Unsupported
```diff
@@ -1,3 +1,3 @@
 [Context]
 shared=unsupported
-unsupported=always
+unsupported=always;if:unsupported-permission:!has-unsupported-permission
```
tchx84, line 3: We're still missing some clarity in this case, e.g., what would happen if the unsupported section used the legacy fallback format `unsupported-permission;if:unsupported-permission:!has-unsupported-permission`?

The outcome would be that `always` and `unsupported-permission` are preserved (which is fine) and only `if:...` will be dropped (which is intended for until we manage to find a simpler way to keep ordering intact for these).

But, we should make our changes in `permissions.js` more obvious for this to be inferred.

src/models/permissions.js
```diff
@@ -192,21 +205,46 @@ var FlatpakPermissionsModel = GObject.registerClass({
...
-                    if (model !== null)
-                        model.loadFromKeyFile(group, key, option, overrides, global);
+                    /* Only the four models in CONDITIONAL_MODELS are
+                     * recognized to support conditionals. A conditional
+                     * entry for anything else is skipped entirely here:
+                     * not loaded into any model, not shown in the UI,
+                     * and never tracked, so nothing is ever written
+                     * back for it either to avoid corrupting the
+                     * override file. */
+                    if (isConditional && !CONDITIONAL_MODELS.includes(model))
```
tchx84, line 239: Since this implementation only supports 1) marking conditionals from original permissions while 2) dropping conditionals from overrides (because it can't guarantee the right order to put it back), I wonder if is still useful to check for `CONDITIONAL_MODELS` here at all.

What could go wrong if we didn't have that now?

src/models/permissions.js
```diff
@@ -192,21 +205,46 @@ var FlatpakPermissionsModel = GObject.registerClass({
...
-                    model = this.constructor._find(`${group}_${key}_${option.replace('!', '')}`);
+                        /* A valid conditional has both an option and a
+                         * condition after "if:". Ignore incomplete
+                         * entries instead of treating them as
+                         * conditionals. */
+                        if (parts.length >= 2) {
+                            isConditional = true;
+                            [bareOption] = parts;
+                        }
+                    }
```
tchx84, line 222: If it happens that we still need `CONDITIONAL_MODELS`, we should make the "dropping" more explicit, and drop it even before we check for any models. For example:

```
/* We drop */
if (isConditional && !CONDITIONAL_MODELS.includes(model))
    return;
```

Additionally, since we went from "completely ignoring conditionals" to only displaying original conditionals, with or without  `CONDITIONAL_MODELS`, it should be clearer that we only will display conditionals from original permissions while we drop conditionals from overrides.

Looks like there's more to analyze here to make sure there aren't any logical gaps.

tests/src/testModels.js
```diff
@@ -630,6 +609,25 @@ describe('Model', function() {
         update();
     });
 
+    it('ignores unsupported conditional permissions', function(done) {
```
tchx84, line 613: Since this implementation is dropping conditionals that aren't original, this test should be replaced/updated to show that.

tchx84 (changes requested):
I think we can still simplify this while making its scope more clear.

tests/content/user/flatpak/overrides/com.test.Unsupported
Malika7188: I can add a comment explaining what happens when the same option appears both as a plain permission and as an unsupported conditional. Or did you have a different implementation in mind?

src/models/permissions.js
Malika7188: > dropping conditionals from overrides (because it can't guarantee the right order to put it back)

Just to clarify this point: currently, it’s the `unsupported conditionals` that get dropped, whether they come from metadata or an override. Conditionals from overrides are marked the same way as the original ones. For example, see the attached screenshot, `devices=kvm;if:kvm:!has-input-device;` comes from an override but is still marked normally just like `sockets_x11` from metadata.


Screenshot From 2026-09-22 07-44-02



---
> I wonder if is still useful to check for `CONDITIONAL_MODELS` here at all.
> 
> What could go wrong if we didn't have that now?

I tested this by removing the check and running the tests, and the app actually crashes with `TypeError: model.markConditional is not a function`. This happens because the model used for unrecognized permissions doesn't have that method.

I also added `markConditional` to `FlatpakUnsupportedModel` to see if that would fix it. It stops the crash, but the unsupported conditional is then saved as a normal permission. So even adding the method does not fix the issue. this check is still needed here.

src/models/permissions.js
Malika7188: > If it happens that we still need `CONDITIONAL_MODELS`, we should make the "dropping" more explicit, and drop it even before we check for any models. For example:
> 
> ```
> /* We drop */
> if (isConditional && !CONDITIONAL_MODELS.includes(model))
>     return;
> ```

the current code already does this. We `check isConditional && !CONDITIONAL_MODELS.includes(model) ` and return before loading or marking the permission.

----
> Additionally, since we went from "completely ignoring conditionals" to only displaying original conditionals, with or without `CONDITIONAL_MODELS`, it should be clearer that we only will display conditionals from original permissions while we drop conditionals from overrides.
> 
> Looks like there's more to analyze here to make sure there aren't any logical gaps.

At the moment, we're dropping unsupported conditionals, not all conditionals from overrides. For example, in the screenshot I shared above, `devices=kvm;if:kvm:!has-input-device` comes from an override but still gets marked.

src/models/permissions.js
Malika7188: @tchx84  apologies I misunderstood your point earlier. I've made the fix. Conditionals coming from overrides (both per-app and global) are no longer marked or shown, only conditionals from the app's own original metadata are.

I have tested this in two ways:

- I added automated test coverage for both per-app and global override scenarios, and all tests pass.
- I've also manually verified using the conditional test app across several cases: conditional only in metadata, conditional in both metadata and an override, and conditional only in an override. The marker behaves correctly in all of them.

PR #887, opened 2026-08-19
models: Fix crash from unreadable or malformed installation configs

Malika7188:
Flatseal reads config files under `installations.d/` to discover extra
Flatpak installation locations (e.g. an SD card). Parsing one of
those files had no error handling, if the file was unreadable or
contained invalid syntax, the exception went uncaught and aborted
Flatseal's startup, before any window opened. No crash dialog, no
visible error, the process just stayed running with nothing shown.

I reproduced it locally both ways (permission-denied file and a
syntactically invalid one) against a real build. Both failed
identically before this fix, and both are handled gracefully after
it.

### DIFF
--- a/src/models/applications.js
+++ b/src/models/applications.js
@@ -122,7 +122,13 @@ var FlatpakApplicationsModel = GObject.registerClass({
         const installations = [];
 
         const keyFile = new GLib.KeyFile();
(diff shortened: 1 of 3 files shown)


Malika7188:
This PR and the related unguarded parsing issues are now addressed in #888

PR #888, opened 2026-08-20
models: Fix crashes caused by malformed or unreadable data

Malika7188:
Flatseal reads data from installed applications and custom Flatpak installations. If some of that data is unreadable or malformed, the resulting error can go uncaught and cause Flatseal to fail silently during startup

This fixes the cases found so far and adds regression tests for them.

### DIFF
--- a/src/models/applications.js
+++ b/src/models/applications.js
@@ -350,9 +350,9 @@ var FlatpakApplicationsModel = GObject.registerClass({
             appdata.version = `${release.get_version()}`;
 
         if (release.get_timestamp() !== null) {
-            const ts = release.get_timestamp();
-            const date = new Date(ts * 1000);
-            appdata.date = date.toISOString().substring(0, 10);
+            const date = GLib.DateTime.new_from_unix_utc(release.get_timestamp());
+            if (date !== null)
+                appdata.date = date.format('%Y-%m-%d');
         }
 
         return appdata;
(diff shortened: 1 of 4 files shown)


src/models/applications.js
```diff
@@ -352,7 +358,8 @@ var FlatpakApplicationsModel = GObject.registerClass({
         if (release.get_timestamp() !== null) {
```
tchx84, line 358: Perhaps we can avoid the issue by rewriting if whole block with:

```
        const date = GLib.DateTime.new_from_unix_utc(release.get_timestamp());
        if (date !== null) {
            appdata.date = date.format("%Y-%m-%d");
        }
```

tests/src/testModels.js
```diff
@@ -757,6 +769,18 @@ describe('Model', function() {
...
 
+    it('does not crash when a custom installation config is malformed', function() {
+        GLib.setenv('FLATPAK_CONFIG_DIR', _customInstallationConfig, true);
+
+        expect(() => applicationsDefault.reload()).not.toThrow();
+
+        const appIds = applicationsDefault.getAll().map(a => a.appId);
+        expect(appIds).toContain(_basicAppId);
+
+        GLib.setenv('FLATPAK_CONFIG_DIR', _flatpakConfig, true);
```
tchx84: This value is only needed for a few tests, and we shouldn't "restore" it here.

tchx84 (changes requested):

tests/src/testModels.js
Malika7188: Hi @tchx84   I've addressed both reviews. timestamp fix uses `GLib.DateTime` now, and `FLATPAK_CONFIG_DIR` resets automatically before every test instead of the test restoring it itself

tchx84 (approved):

